PR #149, opened 2026-04-09
[codex] add binance daily overlay pipeline

mikkokotila:
## What changed

This adds a daily Binance trade overlay in TDW without changing the existing monthly `tdw.binance_trades` ingestion path.

- adds `tdw.binance_daily_trades` as a separate rolling daily table
- adds `tdw.binance_trades_complete` as a plain view that combines final monthly data with the still-open daily overlay without duplicates
- adds cleanup logic so finalized months are removed from the daily overlay after the monthly archive lands
- adds automated Dagster schedules for daily overlay ingestion and monthly roll-forward

## Why

The monthly table remains the trusted canonical source, but it is only complete once Binance publishes the monthly archive. This change fills the gap with official daily archive files so consumers can query a fresher combined view while preserving the existing monthly contract.

## Impact

- `tdw.binance_trades` remains unchanged and continues to represent final monthly data
- `tdw.binance_daily_trades` holds the rolling open-period overlay
- consumers that want the freshest complete picture can query `tdw.binance_trades_complete`
- after deploy, the new table and view need a one-time setup run in Dagster before the schedules fully take over

## Validation

- `python3 -m compileall tdw_control_plane`

### DIFF
--- a/tdw_control_plane/assets/create_binance_trades_complete_view.py
+++ b/tdw_control_plane/assets/create_binance_trades_complete_view.py
@@ -5,8 +5,17 @@ from dagster import asset, AssetExecutionContext
 CLICKHOUSE_HOST = os.environ.get("CLICKHOUSE_HOST", "clickhouse")
 CLICKHOUSE_PORT = int(os.environ.get("CLICKHOUSE_PORT", 9000))
 CLICKHOUSE_USER = os.environ.get("CLICKHOUSE_USER", "default")
-CLICKHOUSE_PASSWORD = os.environ["CLICKHOUSE_PASSWORD"]
+CLICKHOUSE_PASSWORD = os.environ.get("CLICKHOUSE_PASSWORD")
 CLICKHOUSE_DATABASE = os.environ.get("CLICKHOUSE_DATABASE", "tdw")
+
+
+def _get_clickhouse_password():
+    if not CLICKHOUSE_PASSWORD:
+        raise RuntimeError(
+            "CLICKHOUSE_PASSWORD environment variable must be set before creating the ClickHouse client."
+        )
+
+    return CLICKHOUSE_PASSWORD
 
 
 @asset(
@@ -23,7 +32,7 @@ def create_binance_trades_complete_view(context: AssetExecutionContext):
             host=CLICKHOUSE_HOST,
             port=CLICKHOUSE_PORT,
             user=CLICKHOUSE_USER,
-            password=CLICKHOUSE_PASSWORD,
+            password=_get_clickhouse_password(),
             database=CLICKHOUSE_DATABASE,
         )
 
--- a/tdw_control_plane/assets/create_binance_trades_complete_view.py
+++ b/tdw_control_plane/assets/create_binance_trades_complete_view.py
@@ -2,6 +2,13 @@ import os
 from clickhouse_driver import Client as ClickhouseClient
 from dagster import asset, AssetExecutionContext
 
+from tdw_control_plane.assets.create_binance_daily_trades_table import (
+    create_binance_daily_trades_table,
+)
+from tdw_control_plane.assets.create_binance_trades_table import (
+    create_binance_trades_table,
+)
+
 CLICKHOUSE_HOST = os.environ.get("CLICKHOUSE_HOST", "clickhouse")
 CLICKHOUSE_PORT = int(os.environ.get("CLICKHOUSE_PORT", 9000))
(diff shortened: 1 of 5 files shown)


tdw_control_plane/assets/create_binance_trades_complete_view.py
```diff
@@ -0,0 +1,113 @@
...
+CLICKHOUSE_HOST = os.environ.get("CLICKHOUSE_HOST", "clickhouse")
+CLICKHOUSE_PORT = int(os.environ.get("CLICKHOUSE_PORT", 9000))
+CLICKHOUSE_USER = os.environ.get("CLICKHOUSE_USER", "default")
+CLICKHOUSE_PASSWORD = os.environ["CLICKHOUSE_PASSWORD"]
+CLICKHOUSE_DATABASE = os.environ.get("CLICKHOUSE_DATABASE", "tdw")
+
+
+@asset(
+    group_name="tdw_setup",
+    description="Creates a view that combines final monthly Binance trades with open-period daily trades",
```
zero-bang, line 31: The view's SQL references both `binance_trades` and `binance_daily_trades`. ClickHouse validates referenced tables at `CREATE VIEW` time, so if this asset is materialized before both table-creation assets have run, it will fail.

Since all three assets share `group_name="tdw_setup"` with no deps, Dagster may run them in any order (or in parallel).

Fix: add explicit deps so the view is created after both tables:
```python
@asset(
    group_name="tdw_setup",
    deps=[create_binance_daily_trades_table, create_binance_trades_table],
    description="Creates a view that combines final monthly Binance trades with open-period daily trades",
)
```
This requires importing both table-creation assets into this module.

tdw_control_plane/assets/create_binance_trades_complete_view.py
```diff
@@ -0,0 +1,113 @@
...
+                daily.is_buyer_maker,
+                daily.is_best_match,
+                daily.datetime
+            FROM {CLICKHOUSE_DATABASE}.binance_daily_trades AS daily
+            WHERE toStartOfMonth(daily.datetime) > ifNull(
+                (
+                    SELECT toStartOfMonth(max(datetime))
+                    FROM {CLICKHOUSE_DATABASE}.binance_trades
+                ),
+                toDateTime('1969-12-01 00:00:00')
```
zero-bang: `toDateTime('1969-12-01 00:00:00')` is before the Unix epoch. ClickHouse's `DateTime` type is `UInt32` (seconds since 1970-01-01), so pre-epoch values are outside the valid range and the resulting value is version-dependent.

The intent is a sentinel that's older than any real data. Use `toDateTime('1970-01-01 00:00:00')` instead — same effect, stays within the defined range.

tdw_control_plane/definitions.py
```diff
@@ -123,13 +298,94 @@
...
+    run_requests = []
+    current_day = start_day
+    while current_day <= end_date and len(run_requests) < MAX_DAILY_BACKFILL_RUNS_PER_TICK:
+        if current_day not in existing_days:
+            target_date = current_day.isoformat()
+            file_url = (
+                "https://data.binance.vision/data/spot/daily/trades/BTCUSDT/"
+                f"BTCUSDT-trades-{target_date}.zip"
+            )
+            if _binance_archive_available(file_url):
```
zero-bang, line 341: Each iteration of the backfill loop calls `_binance_archive_available()`, which makes an HTTP request with a 30s timeout. With `MAX_DAILY_BACKFILL_RUNS_PER_TICK=31`, a single schedule tick can issue up to 31 HTTP requests (plus multiple ClickHouse round-trips from the helper functions above the loop).

In steady state (1 missing day) this is fine, but during initial deployment or after extended downtime the tick can take a long time and delay other schedule evaluations in the daemon.

This pattern (dynamic evaluation with external I/O) is a better fit for a [Dagster sensor](https://docs.dagster.io/concepts/partitions-schedules-sensors/sensors) than a schedule. As a lighter-weight alternative, consider short-circuiting: if the gap exceeds a small threshold (e.g., 7 days), emit a log and skip — the operator can trigger a manual backfill instead.

zero-bang (changes requested):
Good overall design — the monthly-stays-canonical + daily-overlay approach is sound, the view dedup logic is correct, and the cleanup dependency chain works well.

Three findings below, one of which will cause a hard failure on first deploy.

zero-bang (approved):

zero-bang (approved):

PR #150, opened 2026-04-15
[codex] publish daily BTCUSDT 1m klines to Hugging Face

mikkokotila:
## What changed

This PR adds a TDW-owned Binance spot kline query and a daily Hugging Face publishing path for BTCUSDT 1-minute klines.

- add `tdw_control_plane.query.get_binance_spot_klines` as the canonical spot-kline query backed by `tdw.binance_trades_complete`
- add a partitioned Dagster asset that exports BTCUSDT 1m klines from `2020-01-01` through the latest full UTC day, drops `median` and `iqr`, and uploads the dated Parquet snapshot to Hugging Face
- trigger that export via an asset sensor after successful `insert_daily_binance_trades_to_tdw` materializations
- pass `HF_TOKEN` and `HUGGINGFACE_DATASET_REPO_ID` into the Dagster containers through `docker-compose.yml`
- default the dataset repo to `vaquum/binance_btcusdt_1m_klines`

## Why

We want a TDW-native daily dataset publication flow that does not depend on Limen, while preserving the exact query semantics used today for Binance spot klines.

## Impact

After deploy, the daily TDW update flow can automatically publish snapshots named like `btcusdt_1m_kline_20200101_to_YYYYMMDD.parquet` to Hugging Face, using repo-provided runtime env vars.

## Validation

- `python3 -m py_compile tdw_control_plane/query/__init__.py tdw_control_plane/query/get_binance_spot_klines.py tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py tdw_control_plane/definitions.py`
- `HF_TOKEN=test docker compose config`
- live TDW query check via the existing local tunnel, including writing a sample parquet snapshot for `2026-04-10` with the expected 1440 one-minute rows and the correct dated filename

### DIFF
--- a/tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
+++ b/tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
@@ -118,7 +118,8 @@ def publish_binance_spot_klines_to_huggingface(context: AssetExecutionContext):
         start_date_limit=EXPORT_START_DATE,
         end_date_limit=export_end_exclusive,
         table_name="binance_trades_complete",
-    ).drop(["median", "iqr"])
+        include_quantiles=False,
+    )
 
     if data.height == 0:
         raise RuntimeError(
@@ -166,6 +167,7 @@ def publish_binance_spot_klines_to_huggingface(context: AssetExecutionContext):
             repo_id=dataset_repo_id,
             repo_type="dataset",
             commit_message=commit_message,
+            delete_patterns=[f"{EXPORT_FILENAME_PREFIX}*.parquet"],
         )
 
     latest_datetime = data["datetime"].max()
--- a/tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
+++ b/tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
@@ -1,0 +1,182 @@
+import hashlib
+import json
+import os
+import tempfile
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+
+from dagster import AssetExecutionContext, asset
+from huggingface_hub import HfApi
+
+from tdw_control_plane.assets.daily_trades_to_tdw import daily_partitions
+from tdw_control_plane.query import get_binance_spot_klines
+
+EXPORT_START_DATE = "2020-01-01 00:00:00"
+EXPORT_FILENAME_PREFIX = "btcusdt_1m_kline_20200101_to_"
+
+
+def _get_huggingface_token() -> str:
+    token = os.environ.get("HF_TOKEN") or os.environ.get("HUGGINGFACE_HUB_TOKEN")
+    if not token:
+        raise RuntimeError(
+            "HF_TOKEN or HUGGINGFACE_HUB_TOKEN must be set before publishing to Hugging Face."
+        )
+
+    return token
+
+
+def _get_huggingface_dataset_repo_id() -> str:
+    return os.environ.get(
+        "HUGGINGFACE_DATASET_REPO_ID",
+        "vaquum/binance_btcusdt_1m_klines",
+    )
+
+
+def _build_dataset_card(export_end_date: str, row_count: int, file_name: str) -> str:
+    return f"""# BTCUSDT 1m spot klines
+
+This dataset is exported daily from `tdw.binance_trades_complete` using TDW's `get_binance_spot_klines` query at 1-minute resolution.
+
+Latest snapshot:
+
+- file: `{file_name}`
+- start date: `2020-01-01`
+- rows: `{row_count}`
+- end date: `{export_end_date}`
+- columns: `datetime`, `open`, `high`, `low`, `close`, `mean`, `std`, `volume`, `maker_ratio`, `no_of_trades`, `open_liquidity`, `high_liquidity`, `low_liquidity`, `close_liquidity`, `liquidity_sum`, `maker_volume`, `maker_liquidity`
+
+Notes:
+
+- Source market: Binance spot BTCUSDT
+- Source table: `tdw.binance_trades_complete`
+- `median` and `iqr` are intentionally omitted from the exported Parquet snapshot
+- Timestamps are UTC
+"""
+
+
+def _build_snapshot_metadata(
+    export_end_date: str,
+    file_name: str,
+    row_count: int,
+    file_sha256: str,
+) -> str:
+    generated_at = datetime.now(timezone.utc).isoformat()
+    return (
+        json.dumps(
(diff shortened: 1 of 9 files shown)


tdw_control_plane/query/get_binance_spot_klines.py
```diff
@@ -0,0 +1,105 @@
...
+    if re.fullmatch(r"[A-Za-z0-9_]+", table_name) is None:
+        raise ValueError(f"Invalid ClickHouse table name: {table_name}")
+
+    client = get_clickhouse_client()
+    try:
+        limit = f"LIMIT {n_rows}" if n_rows is not None else ""
+
+        where_clauses = []
+        if start_date_limit is not None:
+            where_clauses.append(f"datetime >= toDateTime('{start_date_limit}')")
```
zero-bang: `start_date_limit` and `end_date_limit` are interpolated directly into SQL via f-strings. The function already validates `table_name` against injection, but the date parameters get no sanitization. Since this is exported as the canonical public API (`__all__`), it should use `clickhouse_connect`'s parameterized query support:

```python
if start_date_limit is not None:
    where_clauses.append("datetime >= toDateTime({start_dt:String})")
    params["start_dt"] = start_date_limit
if end_date_limit is not None:
    where_clauses.append("datetime < toDateTime({end_dt:String})")
    params["end_dt"] = end_date_limit
# ...
arrow_table = client.query_arrow(query, parameters=params)
```

Same applies to `kline_size` and `n_rows` — they're typed as `int` but Python doesn't enforce that at runtime.

tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
```diff
@@ -0,0 +1,182 @@
...
+
+        api = HfApi(token=_get_huggingface_token())
+        api.create_repo(
+            repo_id=dataset_repo_id,
+            repo_type="dataset",
+            exist_ok=True,
+        )
+
+        commit_message = f"Add BTCUSDT 1m klines snapshot through {export_end_date}"
+        api.upload_folder(
```
zero-bang, line 165: Each daily run creates a uniquely-named parquet file (`btcusdt_1m_kline_20200101_to_YYYYMMDD.parquet`) but `upload_folder` never removes previous snapshots. Since every file contains the full dataset from 2020-01-01, this means the HF repo accumulates one redundant full copy per day — unbounded growth.

Pass `delete_patterns` to clean up the previous snapshot:

```python
api.upload_folder(
    folder_path=str(tmp_path),
    repo_id=dataset_repo_id,
    repo_type="dataset",
    commit_message=commit_message,
    delete_patterns=["btcusdt_1m_kline_*.parquet"],
)
```

Alternatively, use a fixed filename (e.g., `btcusdt_1m_klines.parquet`) and let the dated metadata in `latest.json` track provenance.

zero-bang (changes requested):
Good addition — the query semantics look correct and the sensor/job wiring is clean. Two issues to address before merging:

1. **SQL injection in date parameters** — `table_name` is validated but the date strings are interpolated raw into SQL. Since this is the canonical public query function, it should use parameterized queries.
2. **Unbounded snapshot accumulation on HF** — each daily run adds a new uniquely-named parquet file but never removes old ones, so the HF repo grows by one full-dataset copy per day.

Both are straightforward fixes. The rest (sensor plumbing, definitions wiring, dependency additions, dataset card) looks solid.

**Forward-looking concern:** after a multi-day gap, the daily TDW schedule can emit up to 14 `RunRequest`s in a single tick. Each would trigger this sensor, producing 14 sequential full-dataset queries from 2020-01-01 + HF uploads. Worth considering a `minimum_interval_seconds` on the sensor or an incremental export strategy once the dataset grows past a few GB.

mikkokotila:
Addressed the review feedback in `a675a08`.

What changed:
- parameterized the kline query inputs that feed the SQL and added runtime validation for `n_rows`, `kline_size`, and accepted datetime formats
- clarified the `kline_size` units in the public query docstring
- added `include_quantiles` to `get_binance_spot_klines` so the Hugging Face export can skip `median`/`iqr` work entirely
- added `delete_patterns` to the HF upload so the repo keeps only the latest parquet snapshot instead of accumulating a full copy per day

Validation:
- `python3 -m py_compile tdw_control_plane/query/__init__.py tdw_control_plane/query/get_binance_spot_klines.py tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py tdw_control_plane/definitions.py`
- live TDW query check over the tunnel for both the default 19-column query and the slim 17-column export path, plus an invalid-input check for `kline_size=0`

The forward-looking backlog concern is still worth keeping an eye on, but I left that as a follow-up since the two blocking issues are now fixed.

tdw_control_plane/query/get_binance_spot_klines.py
mikkokotila: Fixed in `a675a08`. The query now normalizes accepted datetime formats, validates the integer inputs at runtime, and passes the SQL values through `query_arrow(..., parameters=...)` instead of interpolating the date strings directly.

tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
mikkokotila: Fixed in `a675a08`. The HF upload now uses `delete_patterns` for the parquet snapshot files, so each publish replaces the previous full snapshot instead of leaving one redundant full copy behind per day.

mikkokotila:
@zero-bang the blocking review items are addressed in `a675a08`, the related threads are resolved, and a fresh review request is in. Could you take another look and clear the PR if it now looks good?

zero-bang (approved):

zero-bang (approved):
Verified the fixes in `a675a08` — both blocking items are resolved:

1. **SQL injection** — date parameters, `kline_size`, and `n_rows` all go through ClickHouse parameter binding now (`{start_dt:String}`, `{bucket_seconds:UInt32}`, `{limit_rows:UInt64}`), plus input validation via `_validate_positive_int` and `_normalize_datetime_literal`. Clean.

2. **Snapshot accumulation** — `delete_patterns` on `upload_folder` ensures only the latest parquet file is kept. Exactly what was needed.

The `include_quantiles` flag to skip the `quantileExact` computation on the export path is a nice bonus — saves real ClickHouse CPU time on what will be a growing dataset.

LGTM, good to merge.

zero-bang:
@mikkokotila Both blocking items look good now — approved. The parameterized queries, input validation, and `delete_patterns` on the HF upload are all solid. Ready to merge on your end.

PR #154, opened 2026-04-21
ci: add pr_checks_typing workflow and ratchet-based typing gate

mikkokotila:
## What this gate mechanically enforces

Every rule below is deterministic. The `pr_checks_typing` workflow blocks merge if any single rule is false. Verified on CI against commit `cbccbaf` with pyright 1.1.408 and Python 3.11.

### A note before the rules: what enforces `Any` bans

`[tool.pyright]` sets `reportExplicitAny = "error"` and the gate requires this key to be present with that exact value. **However, pyright 1.1.408 emits `Config contains unrecognized setting "reportExplicitAny"` and does not honor the key at runtime.** The key is kept as a forward-compatible config-presence requirement for the day pyright (or a successor) respects it.

The actual `Any` enforcement today is **sub-gate 4** — an AST walker in `tools/typing_gate.py` that resolves every surface form of `typing.Any` (bare, attribute, aliased import, module-level assignment alias, chained alias) across the package. The regex ratchet in sub-gate 3 is an independent belt-and-suspenders check that catches the common surface forms in textual matches.

---

### Sub-gate 1 — pyright configuration audit (17 rules)

The gate fails the PR if any of these is false.

1. `pyproject.toml` contains a `[tool.pyright]` section.
2. `[tool.pyright].typeCheckingMode` is exactly `"strict"`.
3. `[tool.pyright].reportExplicitAny` is exactly `"error"` (config-presence requirement; see note above).
4. `[tool.pyright].reportMissingImports` is exactly `"error"`.
5. `[tool.pyright].reportMissingTypeStubs` is exactly `"error"`.
6. `[tool.pyright].reportUnknownArgumentType` is exactly `"error"`.
7. `[tool.pyright].reportUnknownMemberType` is exactly `"error"`.
8. `[tool.pyright].reportUnknownVariableType` is exactly `"error"`.
9. `[tool.pyright].reportUnknownLambdaType` is exactly `"error"`.
10. `[tool.pyright].reportUnknownParameterType` is exactly `"error"`.
11. `[tool.pyright].reportMissingParameterType` is exactly `"error"`.
12. `[tool.pyright].reportConstantRedefinition` is exactly `"error"`.
13. `[tool.pyright].reportImportCycles` is exactly `"error"`.
14. `[tool.pyright].include` is exactly `["tdw_control_plane"]` (shrinking this hides files).
15. No key in `[tool.pyright]` whose name starts with `report` is set to any of `"none"`, `"warning"`, `"information"`, `"info"`, `"false"`, or `false`.
16. No `pyrightconfig.json` exists anywhere in the repo outside pruned directories (`.git`, `.venv`, `venv`, `node_modules`, `build`, `dist`, `__pycache__`). Pyright prefers this file over `pyproject.toml`; its presence would silently shadow every rule in 1–15.
17. The workflow invokes the gate with **exactly one of** `--base-budget PATH` or `--bootstrap`, and `--bootstrap` is selected mechanically by the workflow (`git diff --name-only BASE...HEAD` shows `.github/typing_budget.json` was introduced in this commit), not by a human choosing it.

### Sub-gate 2 — budget-source ratchet (10 rules)

The head-ref budget is the PR's own `.github/typing_budget.json`. The base-ref budget is the same file as it exists at the protected branch tip (`origin/main` for PRs / merge-queue, `HEAD~1` for push-to-main). The gate fails the PR if any of these is false when `--base-budget` is passed.

18. `head.package_root` is identical to `base.package_root`.
19. `head.excludes` is a subset of `base.excludes` (head may remove entries; adding an entry hides files from the ratchet).
20. For every pattern key present in `base.patterns`, the same key is present in `head.patterns`.
21. For every pattern key present in both, `head.patterns[key].pattern` is byte-identical to `base.patterns[key].pattern` (the regex cannot be rewritten to one that never matches).
22. For every pattern key present in both, `head.patterns[key].total` is ≤ `base.patterns[key].total`.
23. `head.any_references` is present (the section cannot be deleted).
24. `head.any_references.total` is ≤ `base.any_references.total`.
25. `head.pyright_errors` is present (the section cannot be deleted).
26. `head.pyright_errors.total` is ≤ `base.pyright_errors.total`.
27. If `--bootstrap` is passed instead of `--base-budget`, the workflow has also verified the head commit introduces `.github/typing_budget.json` to the protected ref (otherwise the workflow hard-fails before the gate runs).

### Sub-gate 3 — escape-hatch regex ratchet (10 rules)

Counts are computed across `tdw_control_plane/**/*.py`, excluding entries in `excludes`. Each count compared against the head budget's matching total.

28. Count of `:\s*Any\b` matches is ≤ `patterns.any_annotation.total`.
29. Count of `->\s*Any\b` matches is ≤ `patterns.any_return.total`.
30. Count of `from typing import[^\n]*\bAny\b` matches is ≤ `patterns.any_import.total`.
31. Count of `cast\([^)]*\bAny\b` matches is ≤ `patterns.cast_any.total`.
32. Count of `dict\[[^]]*\bAny\b` matches is ≤ `patterns.dict_any.total`.
33. Count of `list\[\s*Any\b` matches is ≤ `patterns.list_any.total`.
34. Count of `tuple\[[^]]*\bAny\b` matches is ≤ `patterns.tuple_any.total`.
35. Count of `#\s*type:\s*ignore` matches is ≤ `patterns.type_ignore.total`.
36. Count of `#\s*pyright:\s*ignore` matches is ≤ `patterns.pyright_ignore.total`.
37. Count of `#\s*noqa` matches is ≤ `patterns.noqa.total`.

### Sub-gate 4 — AST `Any`-reference ratchet (1 rule; this is where `Any` is actually enforced today)

38. The total number of AST-resolved references to `typing.Any` across `tdw_control_plane/**/*.py` (excluding `excludes`) is ≤ `any_references.total`.

The detector catches every surface form:

- `Any` after `from typing import Any`
- `X` after `from typing import Any as X`
- `typing.Any` after `import typing`
- `t.Any` after `import typing as t`
- Module-level assignment aliases: `A = typing.Any` adds `A` to the `Any` name set.
- Chained aliases: `A = typing.Any; B = A; C = B` — all resolve to the `Any` name set at a fixed point.
- Uses in every position: annotation, return type, generic argument (`list[Any]`, `dict[str, Any]`, `Optional[Any]`), `cast(Any, x)`, variable binding.

Only `ast.Load` contexts are counted as uses, so the LHS of an alias assignment is not double-counted with its RHS.

### Sub-gate 5 — pyright error-count ratchet (1 rule)

39. `summary.errorCount` from `pyright --outputjson`, run on CI (pyright 1.1.408, Python 3.11, `include = ["tdw_control_plane"]`), is ≤ `pyright_errors.total`.

### Sub-gate 6 — `filesAnalyzed` ratchet (1 rule)

40. `summary.filesAnalyzed` from `pyright --outputjson` is ≥ the count of Python files the gate finds under `package_root` with `excludes` applied. A PR that shrinks the analysis surface via `pyright.include`, `pyright.exclude`, or both causes this rule to fail.

---

### Baseline (committed at `.github/typing_budget.json`)

| field | value |
|---|---|
| `package_root` | `"tdw_control_plane"` |
| `excludes` | `["__pycache__", "quickstart_etl_tests", "build", "dist"]` |
| `patterns.*.total` | `0` for every pattern (tdw currently has zero regex-matchable escape hatches) |
| `any_references.total` | `0` |
| `pyright_errors.total` | `1213` |

### How the numbers change

- **Lowering any value in the budget is always allowed.** To lock in an improvement: remove the escape hatches in your PR, run `python tools/typing_gate.py --update-budget --pyright-json pyright_output.json`, commit the updated `.github/typing_budget.json`, push. Sub-gate 2 verifies the new numbers are ≤ base.
- **Raising any value is never allowed** by any mechanism. No flag, no environment variable, no comment syntax, no config toggle. Sub-gate 2 compares against the budget on the protected base ref, fetched by the workflow via `git show BASE:.github/typing_budget.json`. The PR cannot re-oracle itself.
- **Adding a required pyright key** to rules 2–14 requires editing both `tools/typing_gate.py` (`REQUIRED_PYRIGHT`) and `pyproject.toml`. Removing a required key requires only editing the script — which then fails the gate's own config audit on the PR that removes it.
- **Adding a new regex pattern** to the escape-hatch ratchet requires editing `tools/typing_gate.py` (`DEFAULT_PATTERNS`) and regenerating the budget. Sub-gate 2 will require the new key to also be present in future base budgets, locked by the merge of the PR that introduces it.

### Verified transitions

| change | rules that become false | outcome |
|---|---|---|
| unmodified tree | none | PASS |
| add `: Any` to any production file | 28, 38 | FAIL |
| add `typing.Any` to any production file | 38 (regex-blind, AST catches) | FAIL |
| add `t.Any` via `import typing as t` | 38 | FAIL |
| add `A = typing.Any; x: A = 1; y: A = 2` | 38 (counts 3) | FAIL |
| chained alias `A = typing.Any; B = A; C = B; x: C = 1` | 38 (counts 4) | FAIL |
| add `cast(Any, x)` | 31, 38 | FAIL |
| set `reportExplicitAny = "none"` | 3, 15 | FAIL |
| remove `[tool.pyright]` section | 1–14 | FAIL |
| drop `pyrightconfig.json` at repo root | 16 | FAIL |
| shrink `pyright.include` to `[]` | 14, 40 | FAIL |
| narrow head-ref `package_root` | 18 | FAIL |
| add an entry to head-ref `excludes` | 19 | FAIL |
| rewrite a pattern regex in head-ref budget | 21 | FAIL |
| delete a pattern key from head-ref budget | 20 | FAIL |
| raise `pyright_errors.total` in head-ref budget | 26 | FAIL |
| raise `any_references.total` in head-ref budget | 24 | FAIL |
| typecheck regresses by one error | 39 | FAIL |
| pyright `filesAnalyzed` drops below package file count | 40 | FAIL |
| omit `--base-budget` and `--bootstrap` from invocation | 17 | FAIL |
| pass both `--base-budget` and `--bootstrap` | 17 | FAIL |
| `--bootstrap` invoked when head does not introduce the budget | (workflow fails pre-gate) | FAIL |

### Files

- `.github/workflows/pr_checks_typing.yml` — file name, `name:` field, and job name all `pr_checks_typing`.
- `tools/typing_gate.py` — single entry point for all six sub-gates.
- `.github/typing_budget.json` — the committed oracle.
- `pyproject.toml` — `[tool.pyright]` strict + full `report*` matrix, `[tool.ruff]` (E/F/I/UP/RUF/BLE/ANN).
- `.github/workflows/pr_checks_codeql.yml` — bumped `python-version` to `3.11` to match `requires-python`.

### DIFF
--- a/.github/workflows/pr_checks_typing.yml
+++ b/.github/workflows/pr_checks_typing.yml
@@ -1,0 +1,86 @@
+name: pr_checks_typing
+
+# Mechanical typing gate. Three ratchets in one workflow:
+#
+#   1. Pyright configuration must remain strict and must ban explicit Any.
+#      Any weakening (typeCheckingMode != strict, reportExplicitAny != error,
+#      any report* set to none/warning/info/false) fails the build.
+#
+#   2. Escape-hatch patterns are ratcheted against .github/typing_budget.json:
+#      `: Any`, `-> Any`, `cast(Any, ...)`, `dict[..., Any]`, `list[Any]`,
+#      `# type: ignore`, `# pyright: ignore`, `# noqa`. Any file-system-wide
+#      increase fails the build. Decreases may be locked in by updating the
+#      budget file in the same PR.
+#
+#   3. Pyright strict error count is ratcheted against the same budget.
+#      Any increase fails the build. Decreases may be locked in the same way.
+#
+# Philosophy: gates, not language. The gate is what the PR must pass; the
+# documentation in this file is for humans reviewing the gate, not for the
+# agent being gated.
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  merge_group:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  pr_checks_typing:
+    name: pr_checks_typing
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Install build dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e .
+
+      - name: Install pyright
+        run: pip install 'pyright==1.1.409'
+
+      - name: Run pyright strict (capture JSON)
+        run: |
+          set +e
+          pyright --outputjson > pyright_output.json
+          pyright_exit=$?
+          set -e
+          echo "pyright exit code: $pyright_exit"
+          python -c "
+          import json, pathlib, sys
+          p = pathlib.Path('pyright_output.json')
+          if not p.exists() or p.stat().st_size == 0:
+              print('::error::pyright produced no output', file=sys.stderr)
+              sys.exit(2)
+          d = json.loads(p.read_text())
+          s = d.get('summary', {})
+          print(f'pyright: files={s.get(\"filesAnalyzed\")} errors={s.get(\"errorCount\")} warnings={s.get(\"warningCount\")}')
+          "
+
+      - name: Run typing gate
+        run: python tools/typing_gate.py --pyright-json pyright_output.json
+
+      - name: Upload pyright output
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: pyright-output
+          path: pyright_output.json
+          retention-days: 14
--- a/.github/workflows/pr_checks_typing.yml
+++ b/.github/workflows/pr_checks_typing.yml
@@ -99,11 +99,13 @@ jobs:
 
       - name: Run typing gate
         run: |
-          args=( --pyright-json pyright_output.json )
-          if [ -f base_budget.json ]; then
-            args+=( --base-budget base_budget.json )
-          fi
-          python tools/typing_gate.py "${args[@]}"
+          # --base-budget is ALWAYS passed. If the file does not exist
+          # (first commit introducing the gate) the gate will skip the
+          # base-vs-head comparison with a warning. Omitting the flag
+          # would be treated as workflow misconfiguration.
+          python tools/typing_gate.py \
+            --pyright-json pyright_output.json \
+            --base-budget base_budget.json
 
       - name: Upload pyright output
         if: always()
--- a/.github/workflows/pr_checks_typing.yml
+++ b/.github/workflows/pr_checks_typing.yml
@@ -58,12 +58,16 @@ jobs:
       - name: Install pyright
         run: pip install 'pyright==1.1.408'
 
-      - name: Fetch base-ref budget
+      - name: Resolve base ref and fetch base-ref budget
+        id: base
         run: |
-          # The base-ref budget is the protected oracle. The PR's own budget
-          # cannot raise any value against it. For pull_request / merge_group
-          # events the base is github.base_ref; for push-to-main the base is
-          # HEAD~1 (previous tip of main).
+          # The base-ref budget is the protected oracle. If the base has the
+          # file, normal path: fetch it and the gate runs base-vs-head.
+          # If the base does NOT have the file, the only legitimate reason
+          # is that THIS commit adds the file to main for the first time.
+          # The workflow verifies that mechanically by diffing against the
+          # base ref; if the head does not add the file, the gate step
+          # hard-fails. There is no graceful skip.
           set -euo pipefail
           if [ -n "${{ github.base_ref }}" ]; then
             BASE_REV="origin/${{ github.base_ref }}"
@@ -72,11 +76,22 @@ jobs:
             BASE_REV="HEAD~1"
           fi
           echo "base rev: $BASE_REV"
+
           if git show "$BASE_REV:.github/typing_budget.json" > base_budget.json 2>/dev/null; then
+            echo "mode=compare" >> "$GITHUB_OUTPUT"
             echo "base budget fetched ($(wc -c < base_budget.json) bytes)"
           else
-            echo "base rev has no typing_budget.json (first commit introducing the gate)"
+            # Base has no budget. Is THIS commit adding it?
             rm -f base_budget.json
+            # Changed files between base and HEAD
+            CHANGED="$(git diff --name-only "$BASE_REV"...HEAD)"
+            if echo "$CHANGED" | grep -qx '.github/typing_budget.json'; then
+              echo "mode=bootstrap" >> "$GITHUB_OUTPUT"
+              echo "::warning::bootstrap commit: base ref has no typing_budget.json; HEAD introduces it."
+            else
+              echo "::error::base ref has no .github/typing_budget.json and this commit does not introduce it. The budget was deleted from the protected base ref; restore it before merging."
+              exit 1
+            fi
           fi
 
       - name: Run pyright strict (capture JSON)
@@ -99,13 +114,25 @@ jobs:
 
       - name: Run typing gate
         run: |
-          # --base-budget is ALWAYS passed. If the file does not exist
-          # (first commit introducing the gate) the gate will skip the
-          # base-vs-head comparison with a warning. Omitting the flag
-          # would be treated as workflow misconfiguration.
-          python tools/typing_gate.py \
-            --pyright-json pyright_output.json \
-            --base-budget base_budget.json
+          # The previous step decided 'compare' (base ref has the budget)
+          # or 'bootstrap' (this commit introduces it). The two flags are
+          # mutually exclusive; the gate fails if neither is given.
+          case "${{ steps.base.outputs.mode }}" in
+            compare)
+              python tools/typing_gate.py \
+                --pyright-json pyright_output.json \
+                --base-budget base_budget.json
+              ;;
+            bootstrap)
+              python tools/typing_gate.py \
+                --pyright-json pyright_output.json \
+                --bootstrap
+              ;;
+            *)
+              echo "::error::base step did not set mode; refusing to run gate"
+              exit 2
+              ;;
+          esac
 
       - name: Upload pyright output
         if: always()
--- a/.github/workflows/pr_checks_typing.yml
+++ b/.github/workflows/pr_checks_typing.yml
@@ -89,15 +89,22 @@ jobs:
           # is that THIS commit adds the file to main for the first time.
           # The workflow verifies that mechanically by diffing against the
           # base ref; if the head does not add the file, the gate step
-          # hard-fails. There is no graceful skip.
+          # hard-fails. There is no graceful skip and no silent fetch
+          # failure -- if we cannot reach the base ref, the job fails with
+          # the fetch error visible, not on a downstream misbehavior.
           set -euo pipefail
           if [ -n "${{ github.base_ref }}" ]; then
             BASE_REV="origin/${{ github.base_ref }}"
-            git fetch origin "${{ github.base_ref }}" --depth=1 2>/dev/null || true
+            git fetch origin "${{ github.base_ref }}" --depth=1
           else
             BASE_REV="HEAD~1"
           fi
           echo "base rev: $BASE_REV"
+
+          # Verify BASE_REV actually resolves now; if not, the fetch above
+          # produced no usable ref and we must fail loudly rather than
+          # mis-route the downstream compare/bootstrap decision.
+          git rev-parse --verify "$BASE_REV" >/dev/null
 
           if git show "$BASE_REV:.github/typing_budget.json" > base_budget.json 2>/dev/null; then
             echo "mode=compare" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/pr_checks_typing.yml
+++ b/.github/workflows/pr_checks_typing.yml
@@ -54,7 +54,7 @@ jobs:
           pip install -e .
 
       - name: Install pyright
-        run: pip install 'pyright==1.1.409'
+        run: pip install 'pyright==1.1.408'
 
       - name: Run pyright strict (capture JSON)
         run: |
--- a/.github/workflows/pr_checks_typing.yml
+++ b/.github/workflows/pr_checks_typing.yml
@@ -1,23 +1,45 @@
 name: pr_checks_typing
 
-# Mechanical typing gate. Three ratchets in one workflow:
+# Mechanical typing gate. The gate is tools/typing_gate.py; this
+# workflow invokes it. The deterministic rules the gate enforces
+# are listed in full in the PR description; a summary follows.
 #
-#   1. Pyright configuration must remain strict and must ban explicit Any.
-#      Any weakening (typeCheckingMode != strict, reportExplicitAny != error,
-#      any report* set to none/warning/info/false) fails the build.
+#   1. Pyright config audit. Required keys must be present with their
+#      required values (typeCheckingMode=strict, reportExplicitAny,
+#      reportMissingImports, reportMissingTypeStubs, reportUnknown*,
+#      reportMissingParameterType, reportConstantRedefinition,
+#      reportImportCycles -- all "error"). No report* key may be set
+#      to none/warning/info/false. pyright.include must equal
+#      ["tdw_control_plane"]. pyrightconfig.json must not exist
+#      outside vendored paths. Note: pyright 1.1.408 ignores
+#      reportExplicitAny at runtime; the Any enforcement is the AST
+#      ratchet below, not this config key.
 #
-#   2. Escape-hatch patterns are ratcheted against .github/typing_budget.json:
-#      `: Any`, `-> Any`, `cast(Any, ...)`, `dict[..., Any]`, `list[Any]`,
-#      `# type: ignore`, `# pyright: ignore`, `# noqa`. Any file-system-wide
-#      increase fails the build. Decreases may be locked in by updating the
-#      budget file in the same PR.
+#   2. Escape-hatch regex ratchet. Counts `: Any`, `-> Any`,
+#      `cast(Any, ...)`, `dict[..., Any]`, `list[Any]`, `tuple[..., Any]`,
+#      `from typing import ... Any`, `# type: ignore`, `# pyright: ignore`,
+#      `# noqa` across the package. Any file-system-wide increase fails.
 #
-#   3. Pyright strict error count is ratcheted against the same budget.
-#      Any increase fails the build. Decreases may be locked in the same way.
+#   3. AST Any-reference ratchet. Parses each file and counts every
+#      reference to typing.Any: bare `Any`, `typing.Any`, `t.Any`,
+#      aliased imports (`from typing import Any as X`), and module-
+#      level assignment aliases (`A = typing.Any`, chained `B = A`).
+#      Any increase fails. This is the primary Any enforcement today.
 #
-# Philosophy: gates, not language. The gate is what the PR must pass; the
-# documentation in this file is for humans reviewing the gate, not for the
-# agent being gated.
+#   4. Pyright error-count ratchet. summary.errorCount from
+#      `pyright --outputjson` must not increase over base.
+#
+#   5. filesAnalyzed ratchet. pyright must analyze at least as many
+#      files as the gate finds under package_root.
+#
+#   6. Budget-source ratchet. The head-ref budget cannot raise any
+#      number vs the base-ref budget, cannot delete a pattern key,
+#      cannot change package_root, cannot add excludes, and cannot
+#      rewrite a per-pattern regex. Only reductions are allowed.
+#
+# Philosophy: gates, not language. The gate is what the PR must pass.
+# The comment here describes intent for humans reviewing the workflow;
+# the source of truth for rules is tools/typing_gate.py.
 
 on:
   pull_request:
--- a/.github/workflows/pr_checks_typing.yml
+++ b/.github/workflows/pr_checks_typing.yml
@@ -42,6 +42,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # full history, so we can read the budget at base ref
 
       - name: Set up Python
         uses: actions/setup-python@v5
@@ -56,6 +58,27 @@ jobs:
       - name: Install pyright
         run: pip install 'pyright==1.1.408'
 
+      - name: Fetch base-ref budget
+        run: |
+          # The base-ref budget is the protected oracle. The PR's own budget
+          # cannot raise any value against it. For pull_request / merge_group
+          # events the base is github.base_ref; for push-to-main the base is
+          # HEAD~1 (previous tip of main).
+          set -euo pipefail
+          if [ -n "${{ github.base_ref }}" ]; then
+            BASE_REV="origin/${{ github.base_ref }}"
+            git fetch origin "${{ github.base_ref }}" --depth=1 2>/dev/null || true
+          else
+            BASE_REV="HEAD~1"
+          fi
+          echo "base rev: $BASE_REV"
+          if git show "$BASE_REV:.github/typing_budget.json" > base_budget.json 2>/dev/null; then
+            echo "base budget fetched ($(wc -c < base_budget.json) bytes)"
+          else
+            echo "base rev has no typing_budget.json (first commit introducing the gate)"
+            rm -f base_budget.json
+          fi
+
       - name: Run pyright strict (capture JSON)
         run: |
           set +e
@@ -63,7 +86,7 @@ jobs:
           pyright_exit=$?
           set -e
           echo "pyright exit code: $pyright_exit"
-          python -c "
+          python <<'PY'
           import json, pathlib, sys
           p = pathlib.Path('pyright_output.json')
           if not p.exists() or p.stat().st_size == 0:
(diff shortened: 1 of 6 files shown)


tools/typing_gate.py
```diff
@@ -0,0 +1,343 @@
...
+        return tomllib.load(f)
+
+
+def load_budget() -> dict[str, object]:
+    if not BUDGET_PATH.exists():
+        raise SystemExit(
+            f'typing gate cannot run: {BUDGET_PATH.relative_to(REPO_ROOT)} '
+            f'is missing. Run with --update-budget to create it.'
+        )
+    return json.loads(BUDGET_PATH.read_text())
```
mikkokotila, line 876: `load_budget()` reads `.github/typing_budget.json` from the checkout under test, so the PR can raise `pyright_errors.total`, raise any pattern total, or delete a pattern entry and still pass. I reproduced all three in temp checkouts. Exact fix: fetch the budget from the protected base ref in the workflow, pass it to the gate as a separate input, require the candidate budget to preserve the key set, and fail on any increase so only decreases are accepted.

mikkokotila (commented):
GitHub will not let the PR author submit a formal `REQUEST_CHANGES` review on their own pull request, so I am leaving the same requested-changes findings here as inline review comments. The required-status-check issue on `main` is fixed; the comments below cover the remaining enforcement gaps.

tools/typing_gate.py
```diff
@@ -0,0 +1,343 @@
...
+    for key, value in pyright.items():
+        if not isinstance(key, str):
+            continue
+        if key.startswith('report') and value in FORBIDDEN_VALUES:
+            failures.append(
+                f'pyright.{key} = {value!r} -- gate weakening disallowed; '
+                f"must be 'error' (or absent to inherit strict default)"
+            )
+
+    return failures
```
mikkokotila, line 381: This function only audits a handful of `[tool.pyright]` keys; it never asserts the active config source or that pyright is still analyzing `tdw_control_plane`. A PR can add `pyrightconfig.json` (which pyright prefers over `pyproject.toml`) or shrink `include` to a nonexistent path, produce `filesAnalyzed=0`, and still pass this gate. Exact fix: fail if `pyrightconfig.json` exists, require the expected `include` and `exclude` values here, and make the workflow fail if `filesAnalyzed` drops below the base-branch package file count.

tools/typing_gate.py
```diff
@@ -0,0 +1,343 @@
...
+# -------------------------------------------------------------------
+# File walking
+# -------------------------------------------------------------------
+
+def find_python_files(root: Path, excludes: list[str]) -> list[Path]:
+    files: list[Path] = []
+    for p in sorted(root.rglob('*.py')):
+        rel = str(p.relative_to(REPO_ROOT))
+        if any(ex in rel for ex in excludes):
+            continue
```
mikkokotila, line 85: `if any(ex in rel for ex in excludes)` excludes any path that merely contains `build` or `dist`, so a real module like `tdw_control_plane/distributor.py` is skipped entirely. I reproduced that with a temp checkout and the gate passed with `Any` in that file. Exact fix: match exclude entries against path parts or normalized glob patterns rather than raw substrings.

mikkokotila:
Review addressed in `ec98b9e`. CI green on the fix commit.

### Changed

**Path-part excludes (review #3).** `_is_excluded()` now matches path parts, not substrings. An exclude `dist` matches `dist/hello.py` but not `tdw_control_plane/distance_calc.py`. Verified with a four-case assertion.

**Fail-loud file reads (review #4).** `count_pattern()` no longer swallows `OSError` or `UnicodeDecodeError`. Either one now exits the gate with code 2 and a specific `typing_gate: cannot read <file>: <reason>` message. A silently skipped file would under-count escape hatches and let regressions through — the opposite of what this gate is for.

**Budget shape validation (reviews #5 and #6).** Both `gate_escape_hatch_ratchet()` and `gate_pyright_errors()` now validate the JSON shape up front and emit a clean gate failure message instead of a Python traceback when the budget or pyright output is malformed. Applied the smaller of the two variants in the review — guarded `int` coercion with explicit type checks, no 60-line expansion.

**Python-version coherence (reviews #1, #2, #8).** Bumped `requires-python` to `>=3.11`. `pyright.pythonVersion` and `ruff.target-version` were already `3.11`, `tomllib` in `tools/typing_gate.py` requires 3.11+, CI already runs 3.11 — the declared minimum was the only thing out of step. This is the honest alignment; the alternatives (adding a `tomli` fallback or downgrading pyright's target to 3.10) would have preserved the appearance of 3.10 support without the substance.

### Declined

**Heredoc rewrite of the `python -c` block (review #7).** The reviewer's premise is that leading indentation in the `python -c "..."` string produces `IndentationError`. It does not, because the YAML parser strips the block's indent before the command reaches the shell. The step has run successfully in every CI run so far; there is no `IndentationError` in any log. The heredoc would be stylistically equivalent, not a correctness fix, so not changing.

### Verified locally

| scenario | expected | got |
|---|---|---|
| clean tree | PASS | ✓ |
| add `: Any`, `-> Any`, `from typing import Any` to one file | FAIL with three specific pattern messages | ✓ |
| set `any_annotation.total` to `"not a number"` in budget | clean failure message, not traceback | ✓ |
| `_is_excluded('tdw_control_plane/distance_calc.py', ['dist'])` | `False` | ✓ |
| `_is_excluded('dist/hello.py', ['dist'])` | `True` | ✓ |

### Truth-statement delta

The mechanical enforcement rules 1–25 in the PR body are unchanged. What changed:

- The pyright config audit (rules 1–14) is now more robust to malformed `pyproject.toml` — a missing or non-dict `[tool]` no longer crashes.
- The escape-hatch ratchet (rules 15–24) is now more robust to malformed budget JSON — clear failure message in place of traceback.
- The pyright error ratchet (rule 25) is likewise defended against malformed input.

Nothing loosened.

pyproject.toml
```diff
@@ -35,3 +35,79 @@ exclude = ["quickstart_etl_tests*"]
...
+pythonVersion = "3.11"
+typeCheckingMode = "strict"
+include = ["tdw_control_plane"]
+exclude = [
+  "**/__pycache__",
+  "build",
+  "dist",
+  "quickstart_etl_tests",
+]
+reportExplicitAny = "error"
```
mikkokotila, line 66: `reportExplicitAny` still is not mechanically enforced by the pinned toolchain. With `pyright==1.1.408`, pyright emits `Config contains unrecognized setting "reportExplicitAny"` and exits 0, and this workflow does not fail on that warning. Because the escape-hatch ratchet only matches bare `Any`, qualified forms like `typing.Any` still pass; I reproduced `import typing; x: typing.Any = 1` against the current head and the gate returned `TYPING GATE -- PASS`. Exact fix: pin a pyright version whose documented config actually supports explicit-Any reporting and fail the workflow on config warnings/non-zero exit, or replace this claim with an AST-based explicit-Any detector that covers `typing.Any` and aliases.

mikkokotila (commented):
Re-review at `fc49b93`: the original three findings are fixed. One remaining enforcement gap is below.

mikkokotila:
Your review found three real bypasses I had not closed. All three are now mechanically blocked on CI, verified with a nine-scenario local test before pushing. Commits `291a610` + `fc49b93`.

### Your finding A — budget-in-PR bypass

> The gate reads the budget from the PR checkout, so the PR can raise `pyright_errors.total`, raise any pattern total, or delete a pattern entry and still pass.

Closed. The gate now takes a mandatory `--base-budget PATH` argument pointing at the budget as it exists on the protected base ref. New `gate_budget_source()` compares base ↔ head and fails the PR if any of the following is true:

1. `pyright_errors.total` in head > same in base.
2. Any pattern's `total` in head > same in base.
3. A pattern key present in base is absent from head.

The workflow fetches the budget via `git show $BASE_REV:.github/typing_budget.json > base_budget.json` with `$BASE_REV` being `origin/${{ github.base_ref }}` for PRs / merge-queue or `HEAD~1` for push-to-main. For the very first commit introducing the gate (this PR), the base has no budget; the gate prints a warning and skips the base-vs-head check for that one commit only. Every subsequent PR is gated.

### Your finding B — pyrightconfig.json shadow + `filesAnalyzed` sink

> A PR can add `pyrightconfig.json` (pyright prefers it over pyproject.toml) or shrink `include` to a nonexistent path, produce `filesAnalyzed=0`, and still pass.

Two separate closures:

- `gate_pyright_config()` now walks the repo and fails the PR if any `pyrightconfig.json` exists outside vendored dirs (`.venv`, `node_modules`, `build`, `dist`, `.git`). Verified locally — dropping a one-line `{"typeCheckingMode": "off"}` at repo root fails the gate with a specific message.
- `gate_pyright_config()` now also requires `pyright.include` to equal `["tdw_control_plane"]` exactly. Any deviation fails. Verified locally — setting `include = []` fails with "shrinks the analysis surface."
- `gate_files_analyzed()`: new. After pyright runs, the gate reads `summary.filesAnalyzed` and compares it to `len(find_python_files(package_root, excludes))`. If pyright analyzed fewer files than the package contains, the PR fails. Verified locally — feeding a synthetic pyright-output with `filesAnalyzed=0` fails with "errors are hidden outside the analyzed set."

### Your finding C — substring excludes

> A real module like `tdw_control_plane/distributor.py` is skipped entirely; I reproduced it.

`ec98b9e` (from the earlier round) switched `find_python_files` to path-part matching. Re-tested the specific scenario you described:

```
_is_excluded('tdw_control_plane/distributor.py', ['dist']) -> False
_is_excluded('dist/hello.py',                     ['dist']) -> True
_is_excluded('tdw_control_plane/builds_meta.py',  ['build']) -> False
_is_excluded('build/x.py',                        ['build']) -> True
```

The fix landed one commit before your review was posted, so the temp-checkout scenario you reproduced is already closed. Confirmed here because the review asks for explicit reproduction.

### Nine-scenario local verification run before pushing

| # | scenario | expected | got |
|---:|---|---|---|
| 1 | clean tree, base-budget provided | PASS | PASS |
| 2 | `--base-budget` flag omitted | FAIL (misconfig) | FAIL with specific message |
| 3 | `--base-budget` path missing on disk | SKIP with warning | SKIP with warning |
| 4 | PR raises `pyright_errors.total` | FAIL | FAIL naming base / head numbers |
| 5 | PR deletes a pattern key | FAIL | FAIL naming the deleted key |
| 6 | PR raises any pattern total | FAIL | FAIL naming the key and deltas |
| 7 | PR adds `pyrightconfig.json` at root | FAIL | FAIL with delete-this-file message |
| 8 | PR shrinks `pyright.include` to `[]` | FAIL | FAIL with "shrinks analysis surface" |
| 9 | pyright `filesAnalyzed=0` in output | FAIL | FAIL naming expected file count |

### CI state

Both workflows green on the head commit (`fc49b93`):

- `pr_checks_typing`: SUCCESS (47 s)
- `CodeQL (python)`: SUCCESS (the earlier failure was the `requires-python` bump cascading into a Python-version mismatch on the CodeQL workflow; bumped it to 3.11 alongside)

### Truth-statement delta (updated from the earlier 25)

The PR body's 25-rule enforcement list now has five additions covering your findings. I'll update the PR body next message with the expanded list.

mikkokotila:
Reproduced your scenario first: appended `import typing; _regression: typing.Any = 1` to a production file on head, ran the gate, got `TYPING GATE -- PASS`. Confirmed the hole. Closed in `8c2c782`.

### What changed

**1. Removed the aspirational claim.** `reportExplicitAny` is no longer in `REQUIRED_PYRIGHT`. The pinned `pyright==1.1.408` emits `Config contains unrecognized setting "reportExplicitAny"` and proceeds at exit 0; the setting has no mechanical effect on this toolchain. The key still sits in `pyproject.toml` (forward-compatible if a later pyright respects it), but the gate no longer asserts enforcement the toolchain does not provide. That's exactly the Origo pattern I should not be reproducing — documentation standing in for code.

**2. Replaced regex Any-detection with AST.** New `count_any_references_ast()` in `tools/typing_gate.py`. For each production file:

- Parses with `ast.parse`.
- Walks `ast.ImportFrom` and `ast.Import` to build two sets of local names: names that directly bind to `typing.Any` (from `from typing import Any`, `from typing import Any as A`), and names that bind to the `typing` module (`import typing`, `import typing as t`).
- Walks every other node. Counts any `ast.Name` whose `.id` is in the direct set, and any `ast.Attribute` whose `.attr == 'Any'` and whose value is a `Name` in the module-alias set.

What this catches that the regex missed:

| surface form | regex `:\s*Any\b` | AST |
|---|---|---|
| `x: Any` | ✓ | ✓ |
| `x: typing.Any` | ✗ | ✓ |
| `x: t.Any` after `import typing as t` | ✗ | ✓ |
| `x: A` after `from typing import Any as A` | ✗ | ✓ |
| `dict[str, typing.Any]` | ✗ | ✓ |
| `cast(typing.Any, x)` | ✗ | ✓ |
| `list[A]` (after alias) | ✗ | ✓ |

**3. New sub-gate.** `gate_any_references_ast(budget)` ratchets the AST count against `.github/typing_budget.json['any_references']['total']` (baseline captured on tdw: 0). Any increase fails merge.

**4. Budget schema v2.** Added top-level `any_references: {"total": N}`. `--update-budget` auto-migrates v1 → v2. The base-vs-head ratchet (`gate_budget_source`) treats `any_references` like `pyright_errors` — the head budget cannot raise its own total or delete the section.

### Verified locally

Each scenario adds the stated code to one production file, then runs the gate against the committed baseline. The regex ratchet and the AST ratchet are independent; where both fire, they're listed separately in the output.

| # | scenario | regex-gate | AST-gate | verdict |
|---:|---|---|---|---|
| 1 | clean tree | — | — | PASS |
| 2 | `import typing; x: typing.Any = 1` | miss | **catch** | FAIL |
| 3 | `from typing import Any as A; x: A = 1` | catch (import) | **catch** | FAIL |
| 4 | `import typing as t; x: t.Any = 1` | miss | **catch** | FAIL |
| 5 | `cast(typing.Any, 1)` | catch | **catch** | FAIL |
| 6 | `dict[str, typing.Any]` | catch | **catch** | FAIL |

Scenario 2 is the exact reproduction from your review. Before this commit: `TYPING GATE -- PASS`. After this commit:

```
TYPING GATE -- FAIL
  gate: any-references-ast-ratchet
    - any_references (AST): budget=0 current=1 (delta=1). Covers Any, typing.Any, t.Any, aliased imports. Remove the new Any or lower the budget.
```

### CI

- `pr_checks_typing` ✓ SUCCESS on `8c2c782`. The bootstrap annotation still fires (expected — this PR introduces the budget file; next PR onward will run the base-vs-head path).
- `CodeQL (python)` ✓ SUCCESS.

.github/workflows/pr_checks_typing.yml
```diff
@@ -0,0 +1,143 @@
+name: pr_checks_typing
+
+# Mechanical typing gate. Three ratchets in one workflow:
+#
+#   1. Pyright configuration must remain strict and must ban explicit Any.
+#      Any weakening (typeCheckingMode != strict, reportExplicitAny != error,
+#      any report* set to none/warning/info/false) fails the build.
```
mikkokotila: This comment is no longer true as written. The current gate does **not** require `reportExplicitAny = "error"` to remain present: removing that key now passes, while setting it to `none` fails only because the generic `report*` weakening check catches forbidden values. I reproed both against the current head. The implementation now enforces explicit-`Any` primarily through the AST ratchet (`any_references`), not through exact-key pyright config enforcement, so this header comment and the PR body should be updated to match the actual mechanism.

mikkokotila (commented):
Re-review at `8c2c782`: the enforcement bypasses appear fixed. One remaining user-visible claim is stale.

mikkokotila:
Reproduced all five current threads before fixing. Three were real bypasses, two were stale-doc regressions I introduced in the previous commit. All eight scenarios now behave correctly. `cbccbaf`.

### Your thread (workflow:7) — stale docs

You were right. Removing `reportExplicitAny` from pyproject.toml passed the gate on `8c2c782`, while every doc claimed it was enforced. Fixed by bringing `reportExplicitAny` back into `REQUIRED_PYRIGHT` as a **config-presence check** and making every doc honest about what does what:

- `REQUIRED_PYRIGHT` comment in `tools/typing_gate.py`: enumerates which keys pyright actually enforces at runtime and explicitly calls out that `reportExplicitAny` is ignored by pyright 1.1.408 — kept in the list for forward-compatibility, with the AST ratchet doing the real Any enforcement today.
- Workflow comment in `pr_checks_typing.yml`: rewritten to describe the six sub-gates, with the same note about `reportExplicitAny` / AST.
- `pyproject.toml` comment block: same note.

Verified: removing `reportExplicitAny = "error"` from `pyproject.toml` now fails the gate with `pyright.reportExplicitAny must be 'error', got None`.

### Copilot thread (typing_gate.py:146) — AST bypassed by assignment aliases

Reproduced:

```python
import typing
A = typing.Any
_r1: A = 1   # not counted before
_r2: A = 2   # not counted before
... 5 uses ...
```

On `8c2c782`, the gate saw `current=1` (the `typing.Any` on the RHS of `A = typing.Any`). Five uses of `: A` were invisible.

Fix: `_collect_any_bindings` now has a second pass that walks `ast.Assign` / `ast.AnnAssign`, and when the RHS resolves to `typing.Any`, adds LHS names to the `direct` set. Iterates to a fixed point (bounded at 16 rounds) so chained aliases resolve: `A = typing.Any; B = A; C = B` all end up in `direct`.

`_is_any_reference` now also restricts `ast.Name` matches to `ast.Load` context, so the LHS of `A = typing.Any` doesn't get double-counted as a use.

Verified:

| input | before | after |
|---|---:|---:|
| `A = typing.Any` + 5 uses of `: A` | 1 | **6** |
| `A = typing.Any; B = A; C = B` + 1 use of `: C` | 1 | **4** |
| `from typing import Any` + 1 use of `: Any` | 1 | 1 (no double count) |

### Copilot thread (typing_gate.py:567) — budget-field tampering

Reproduced three bypasses:

1. **Narrow `package_root`**: budget had `"package_root": "tdw_control_plane"`. Set it to `"tdw_control_plane/utils"` in the head budget. Gate **passed** — the scan now covers only the 1-file `utils/` subdir.
2. **Add an exclude**: add `"tdw_control_plane/assets"` to `excludes`. Gate **passed** — the 5 files under `assets/` are no longer scanned for Any.
3. **Weaken a regex**: rewrite `any_annotation.pattern` from `':\\s*Any\\b'` to `'__NEVER_MATCH__'`. Gate **passed** — that ratchet key is now neutralized even though its `total` stays at 0.

Fix: `gate_budget_source` now additionally asserts:

- `package_root` **identical** between base and head.
- `excludes` in head is a **subset** of excludes in base (can remove, cannot add).
- Each pattern's `pattern` regex string **identical** for any key present in base.

All three bypasses now fail with specific messages.

### Copilot threads (gate_pyright_errors:406, gate_files_analyzed:617) — OSError

Correct point. `path.read_text()` only caught `JSONDecodeError`; an unreadable file (permissions, transient FS error) would raise `OSError` and crash the script with a traceback instead of returning the controlled gate failure. Fixed both functions — they now wrap the read in `try/except OSError` and return a clean gate failure message if the read fails.

### Copilot thread (typing_gate.py:256) — `.git` traversal

Correct. `REPO_ROOT.rglob('pyrightconfig.json')` descends into `.git`, which on a `fetch-depth: 0` CI checkout is sizable. Replaced with `os.walk` that prunes `dirnames` in place at each level, removing `.git`, `.venv`, `venv`, `node_modules`, `build`, `dist`, `__pycache__` before recursing. Cleaner and faster; no pyrightconfig.json inside those paths would shadow our config anyway.

### Eight-scenario local verification

| # | scenario | expected | got |
|---:|---|---|---|
| 1 | clean tree, base provided | PASS | PASS |
| 2 | `A = typing.Any` + 5 uses of `: A` | FAIL | FAIL (6 AST counts) |
| 3 | `A = typing.Any; B = A; C = B` + 1 use | FAIL | FAIL (4 AST counts) |
| 4 | narrow `package_root` in head budget | FAIL | FAIL (budget-source ratchet) |
| 5 | add exclude in head budget | FAIL | FAIL (budget-source ratchet) |
| 6 | rewrite a pattern regex | FAIL | FAIL (budget-source ratchet) |
| 7 | remove `reportExplicitAny` from pyproject | FAIL | FAIL (pyright-config) |
| 8 | normal `from typing import Any; x: Any` | 1 AST count (no double-count) | ✓ |

### CI

- `pr_checks_typing` ✓ SUCCESS on `cbccbaf`. Bootstrap annotation still fires as expected (this PR introduces the budget file).
- `CodeQL (python)` ✓ SUCCESS.

### Outstanding threads that are OUTDATED

Five threads are marked `isOutdated: true` in the API because the commit they were filed against no longer exists in the head diff. All five reference issues I've since fixed, so they're already addressed by the cumulative commit history. I'm not resolving them explicitly — happy to if you prefer, let me know.

mikkokotila (commented):
Re-review at `cbccbaf`: I did not reproduce any remaining enforcement bypass in the current code. The remaining issue is the PR body: it is now stale and contains at least one materially false statement about what the gate does.

mikkokotila:
Both threads closed in `fafe6a9`.

### Thread on `.github/workflows/pr_checks_typing.yml:96` (base-ref fetch)

`git fetch origin "${{ github.base_ref }}" --depth=1 2>/dev/null || true` masked two things: (a) the fetch failing, and (b) the reason it failed. If the fetch silently failed, the later `git show` would also fail (incorrectly routing into bootstrap or dropping into the "budget deleted" branch), and neither error message would reference the real cause.

Fix: removed both the `|| true` and the `2>/dev/null`. Added a `git rev-parse --verify "$BASE_REV"` immediately after the fetch, so if the base ref cannot be resolved, the job fails at that line with the fetch error visible — not downstream.

### Thread on `tools/typing_gate.py:885` (optional `--pyright-json`)

Real bypass. The pyright-error and filesAnalyzed ratchets (sub-gates 5 and 6, rules 39 and 40) are keyed off the pyright JSON. With the flag optional, a workflow edit that dropped `--pyright-json pyright_output.json` would silently skip both.

Fix: `--pyright-json` is now required on normal gate runs. Omitting it fails the gate with:

```
TYPING GATE -- FAIL

  gate: runtime-args
    - --pyright-json was not provided. Normal gate runs require pyright
      output so the pyright-error-ratchet and files-analyzed-ratchet can
      run. Pass --pyright-json PATH or, if regenerating the budget, use
      --update-budget instead.

1 failure(s). Merge blocked.
```

`--update-budget` is the one legitimate exception (it regenerates the budget and can run without pyright output) — preserved.

### Verified

| scenario | expected | got |
|---|---|---|
| normal: `--pyright-json` + `--base-budget` | PASS | PASS |
| `--pyright-json` omitted | FAIL with `runtime-args` message | FAIL with runtime-args message |
| `--update-budget` without `--pyright-json` | still works | still works |
| base ref unreachable on CI (simulated with bogus ref) | workflow fails at `git rev-parse` | confirmed in separate local test |

### CI

- `pr_checks_typing` ✓ SUCCESS on `fafe6a9`.
- `CodeQL (python)` running; will clear shortly.

Issue #157, opened 2026-04-21
feat: add Origo daily spot trades raw table, ingestion ledger, and pr_checks_tests gate


mikkokotila:
## Slice ID

S02

## Slice Title

feat: add Origo daily spot trades raw table, ingestion ledger, and pr_checks_tests gate

## Routing

> **Significance.** Routing is the minimum information needed to place this
> slice in the dependency graph. Parent and dependencies must be existing
> issue numbers; unlocks may be forward-looking. Prose description of the
> relationship is not accepted — only issue references. This section is not
> a planning document; it is a graph edge list.

Parent issue: none (foundational Origo source-native template)
Depends on:
- #155
Unlocks:
- Deterministic replication of this daily-source-native pattern to other Binance daily datasets
- Deterministic replication of this daily-source-native pattern to other exchange datasets
- Later widening of `.github/workflows/pr_checks_tests.yml` beyond the Origo path

## Minimum Viable Capability

> **Significance.** A slice delivers exactly one capability. The capability
> must be expressible as a set of assertions that become mechanically true
> at merge. This is the answer to *what changes in the system that an
> external observer can verify by running a command?* Every line below
> must be evaluable as a shell command, SQL query, Python expression, or
> HTTP request that returns a definite result. One assertion per line.
>
> Accepted:
> - `clickhouse-client --query "SELECT countIf(stream_sequence IS NULL) FROM canonical_event_log"` returns `0`.
> - `python -c "import inspect, origo.events.writer as m; print(inspect.signature(m.CanonicalEventWriter.__init__))"` returns `(self, *, clickhouse_client)`.
> - `curl -sf https://origo.vaquum.fi/v1/health/schema | jq -r '.canonical_event_log_v1.columns | length'` returns `17`.
>
> Rejected:
> - *Freeze the current-truth read contract.*
> - *Strict CQRS separation.*
> - *Terminal proof obligations are complete.*

Assertions (one per line, each executable as-is):
- `pytest tests/origo_source_native/test_runtime_contract.py -q -k "dockerfile_uses_python_311"` exits `0`
- `pytest tests/origo_source_native/test_origo_daily_spot_trades.py -q -k "create_origo_database_is_idempotent"` exits `0`
- `pytest tests/origo_source_native/test_origo_daily_spot_trades.py -q -k "create_binance_daily_spot_trades_table_is_idempotent"` exits `0`
[...]

PR #161, opened 2026-04-22
ci: enforce required PR-path checks and ruleset drift on main

### LINKED ISSUE #160: ci: enforce required PR-path checks and ruleset drift on main
## Slice ID

S03

## Slice Title

ci: enforce required PR-path checks and ruleset drift on main

## Routing

Parent issue: none
Depends on:
- #155
Unlocks:
- Full blocking enforcement of the repository's current required PR-path checks on `main`, plus `pr_checks_ruleset` drift detection
- A checked-in ruleset snapshot with loud drift detection instead of ad-hoc API mutation
- Preservation of `strict_required_status_checks_policy: true` as audited live ruleset state
- Retirement of the non-authoritative post-merge CHANGELOG/version updater in favor of pre-merge enforcement by `pr_checks_version`
- Linked-issue title validation in `pr_checks_cc` that fails loud instead of silently skipping on GitHub API failure

## Minimum Viable Capability

>
> Accepted:
>
> Rejected:
> - *Strict CQRS separation.*

- `python3 tools/ruleset_gate.py --live-json tests/fixtures/github/ruleset_live_target.json --ruleset-file .github/rulesets/main.json --ruleset-id 5406599` exits `0`
- `python3 tools/ruleset_gate.py --live-json tests/fixtures/github/ruleset_live_target_without_bypass_actors.json --ruleset-file .github/rulesets/main.json --ruleset-id 5406599` exits `0`
- `python3 tools/ruleset_gate.py --live-json tests/fixtures/github/ruleset_live_with_bypass_actor.json --ruleset-file .github/rulesets/main.json --ruleset-id 5406599` exits `1`
- `python3 tools/ruleset_gate.py --live-json tests/fixtures/github/ruleset_live_unexpected_field.json --ruleset-file .github/rulesets/main.json --ruleset-id 5406599` exits `1`
- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[] | select(.type=="required_status_checks") | .parameters.strict_required_status_checks_policy'` returns `true`
- `rg -n 'str \| None|return None|if issue_title is not None|or None if the issue cannot be' tools/cc_gate.py | wc -l | tr -d ' '` returns `0`
- `test -f .github/workflows/pr_post_changelog.yml; echo $?` returns `1`
- `test -f scripts/update_changelog.py; echo $?` returns `1`

## Current State

>
> Accepted:
>
> Rejected:

- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[].type | @json'` returns the 5-line set `"deletion"`, `"non_fast_forward"`, `"pull_request"`, `"copilot_code_review"`, `"required_status_checks"`
- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[] | select(.type=="required_status_checks") | .parameters.strict_required_status_checks_policy'` returns `true`
- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context | @json'` returns the 6-line set `"PR Checks CodeQL (python)"`, `"pr_checks_typing"`, `"pr_checks_cc"`, `"pr_checks_slice"`, `"pr_checks_fail_loud"`, `"pr_checks_version"`
- `test -f .github/rulesets/main.json; echo $?` returns `1`
- `test -f .github/workflows/pr_checks_ruleset.yml; echo $?` returns `1`
- `test -f tools/ruleset_gate.py; echo $?` returns `1`
- `gh workflow list --repo Vaquum/tdw-control-plane | rg '^pr_checks_ruleset$'` exits `1`
- `gh workflow list --repo Vaquum/tdw-control-plane | rg '^Post-Merge CHANGELOG Update'` returns `Post-Merge CHANGELOG Update\tactive\t231836243`
- `test -f scripts/update_changelog.py; echo $?` returns `0`
- `rg -n 'str \| None|return None|if issue_title is not None|or None if the issue cannot be' tools/cc_gate.py | wc -l | tr -d ' '` returns `6`

## Target State

> predicate.
>
> Accepted:

- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[].type | @json'` returns the 5-line set `"deletion"`, `"non_fast_forward"`, `"pull_request"`, `"copilot_code_review"`, `"required_status_checks"`
- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[] | select(.type=="required_status_checks") | .parameters.strict_required_status_checks_policy'` returns `true`
- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context | @json'` returns the 7-line set `"PR Checks CodeQL (python)"`, `"pr_checks_typing"`, `"pr_checks_cc"`, `"pr_checks_slice"`, `"pr_checks_fail_loud"`, `"pr_checks_version"`, `"pr_checks_ruleset"`
- `python3 tools/ruleset_gate.py --ruleset-file .github/rulesets/main.json --repo Vaquum/tdw-control-plane --ruleset-id 5406599` exits `0`
- `gh workflow list --repo Vaquum/tdw-control-plane | rg '^pr_checks_ruleset$'` returns one active workflow line
- `gh workflow list --repo Vaquum/tdw-control-plane | rg '^Post-Merge CHANGELOG Update'` exits `1`
- `test -f scripts/update_changelog.py; echo $?` returns `1`
- `rg -n 'str \| None|return None|if issue_title is not None|or None if the issue cannot be' tools/cc_gate.py | wc -l | tr -d ' '` returns `0`

## Surfaces

- `CHANGELOG.md`
- `pyproject.toml`
- `tools/cc_gate.py`

Added:
- `tools/ruleset_gate.py`
- `tests/tools/test_cc_gate.py`
- `tests/tools/test_ci_contract.py`
- `tests/fixtures/github/ruleset_live_unexpected_field.json`

Removed:
- `.github/workflows/pr_post_changelog.yml`
- `scripts/update_changelog.py`

## Signature Diff

### .github/rulesets/main.json::rules
Before:
```
(file does not exist)
```
After:
```json
{
  "name": "Protect-Main",
  "target": "branch",
  "enforcement": "active",
  "conditions": {
    "ref_name": {
      "exclude": [],
      "include": ["refs/heads/main"]
    }
  },
  "bypass_actors": [],
  "rules": [
    {"type": "deletion"},
    {"type": "non_fast_forward"},
    {
      "type": "pull_request",
      "parameters": {
        "required_approving_review_count": 0,
        "dismiss_stale_reviews_on_push": false,
        "required_reviewers": [],
        "require_code_owner_review": false,
        "require_last_push_approval": false,
        "required_review_thread_resolution": true,
        "allowed_merge_methods": ["merge", "squash", "rebase"]
      }
    },
    {
      "type": "copilot_code_review",
      "parameters": {
        "review_on_push": true,
        "review_draft_pull_requests": false
      }
    },
    {
      "type": "required_status_checks",
      "parameters": {
        "strict_required_status_checks_policy": true,
        "do_not_enforce_on_create": false,
        "required_status_checks": [
          {"context": "PR Checks CodeQL (python)", "integration_id": 15368},
          {"context": "pr_checks_typing", "integration_id": 15368},
          {"context": "pr_checks_cc", "integration_id": 15368},
          {"context": "pr_checks_slice", "integration_id": 15368},
          {"context": "pr_checks_fail_loud", "integration_id": 15368},
          {"context": "pr_checks_version", "integration_id": 15368},
          {"context": "pr_checks_ruleset", "integration_id": 15368}
        ]
      }
    }
  ]
}
```

### .github/workflows/pr_checks_ruleset.yml::workflow
Before:
```
(file does not exist)
```
After:
```yaml
name: pr_checks_ruleset
on:
  pull_request:
    branches: [main]
  workflow_dispatch:
permissions:
  contents: read
jobs:
  pr_checks_ruleset:
    name: pr_checks_ruleset
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v4
      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version: '3.11'
      - name: Install test dependency
        run: python -m pip install --upgrade pip pytest
      - name: Run CI contract tests
        run: pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py -q
      - name: Compare live ruleset to checked-in snapshot
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: python3 tools/ruleset_gate.py --ruleset-file .github/rulesets/main.json --repo "${{ github.repository }}" --ruleset-id 5406599
```

### tools/ruleset_gate.py::normalize_snapshot_ruleset + normalize_live_ruleset + main
Before:
```
(file does not exist)
```
After:
```python
REQUIRED_TOP_LEVEL_FIELDS = frozenset({
    'name',
    'target',
    'enforcement',
    'conditions',
    'rules',
})
OPTIONAL_LIVE_TOP_LEVEL_FIELDS = frozenset({'bypass_actors'})
SNAPSHOT_TOP_LEVEL_FIELDS = frozenset(
    REQUIRED_TOP_LEVEL_FIELDS | OPTIONAL_LIVE_TOP_LEVEL_FIELDS
)
IGNORED_LIVE_FIELDS = frozenset({
    'current_user_can_bypass',
    'node_id',
    'created_at',
    'updated_at',
    '_links',
    'id',
    'source',
    'source_type',
})

def normalize_snapshot_ruleset(payload: dict[str, Any]) -> dict[str, Any]:
    missing = SNAPSHOT_TOP_LEVEL_FIELDS - set(payload)
    if missing:
        raise SystemExit(
            f'expected snapshot ruleset field(s) missing: {sorted(missing)}'
        )
    return payload

def normalize_live_ruleset(payload: dict[str, Any]) -> tuple[dict[str, Any], frozenset[str]]:
    unexpected = (
        set(payload)
        - REQUIRED_TOP_LEVEL_FIELDS
        - OPTIONAL_LIVE_TOP_LEVEL_FIELDS
        - IGNORED_LIVE_FIELDS
    )
    if unexpected:
        raise SystemExit(
            f'unexpected live ruleset field(s): {sorted(unexpected)}'
        )
    comparable_fields = REQUIRED_TOP_LEVEL_FIELDS | (
        OPTIONAL_LIVE_TOP_LEVEL_FIELDS & set(payload)
    )
    missing_optional = OPTIONAL_LIVE_TOP_LEVEL_FIELDS - set(payload)
    if missing_optional:
        print(
            'live ruleset omitted non-observable field(s): ' 
            f'{sorted(missing_optional)}; comparing observable subset only',
            file=sys.stderr,
        )
    return (
        {key: payload[key] for key in sorted(comparable_fields)},
        frozenset(comparable_fields),
    )

def main() -> int:
    parser = argparse.ArgumentParser(description='Ruleset drift gate')
    parser.add_argument('--ruleset-file', required=True)
    parser.add_argument('--repo')
    parser.add_argument('--ruleset-id', required=True)
    parser.add_argument('--live-json')
    ...
    expected_ruleset = normalize_snapshot_ruleset(
        load_json_file(Path(args.ruleset_file))
    )
    live_ruleset, compared_fields = normalize_live_ruleset(load_live_ruleset(...))
    expected_ruleset = {key: expected_ruleset[key] for key in sorted(compared_fields)}
    if live_ruleset != expected_ruleset:
        print('ruleset drift detected', file=sys.stderr)
        return 1
    return 0
```

### tools/cc_gate.py::fetch_issue_title
Before:
```python
def fetch_issue_title(repo: str, number: int) -> str | None:
    """Return the title of the issue, or None if the issue cannot be
    resolved (slice_gate is the authoritative check for existence /
    openness / labels; cc_gate only reads the title)."""
    ...
    if result.returncode != 0:
        return None
```
After:
```python
def fetch_issue_title(repo: str, number: int) -> str:
    """Return the title of the issue. Any gh failure is a gate setup
    failure because linked-issue title validation is part of this gate's
    own contract."""
    ...
    if result.returncode != 0:
        print(..., file=sys.stderr)
        raise SystemExit(2)
```

### tools/cc_gate.py::gate
Before:
```python
    refs = find_closing_references(pr_body)
    if len(refs) == 1:
        issue_title = fetch_issue_title(repo, refs[0])
        if issue_title is not None:
            err = check_cc(issue_title)
```
After:
```python
    refs = find_closing_references(pr_body)
    if len(refs) == 1:
        issue_title = fetch_issue_title(repo, refs[0])
        err = check_cc(issue_title)
```

### .github/workflows/pr_post_changelog.yml::name
Before:
```yaml
name: Post-Merge CHANGELOG Update
```
After:
```
(file removed)
```

### scripts/update_changelog.py::shebang
Before:
```python
#!/usr/bin/env python3
```
After:
```
(file removed)
```

## Tests That Prove the MVC

|---|---|---|
| ruleset snapshot matches complete target live JSON after stripping only the named API-only live fields | `tests/tools/test_ruleset_gate.py` | `test_matches_target_snapshot` |
| repo-hosted live reads that omit `bypass_actors` still compare the observable subset loudly | `tests/tools/test_ruleset_gate.py` | `test_missing_bypass_actors_is_allowed_for_repo_hosted_reads` |
| bypass actor drift is treated as failure when the field is observable | `tests/tools/test_ruleset_gate.py` | `test_bypass_actor_drift_is_failure` |
| unexpected top-level live ruleset field is treated as drift | `tests/tools/test_ruleset_gate.py` | `test_unexpected_top_level_field_in_live_is_drift` |
| `cc_gate` no longer carries nullable linked-issue fetch semantics | `tests/tools/test_cc_gate.py` | `test_issue_title_fetch_is_non_nullable_and_gate_has_no_none_guard` |
| post-merge changelog workflow is retired | `tests/tools/test_ci_contract.py` | `test_post_merge_changelog_workflow_removed` |
| post-merge changelog script is retired | `tests/tools/test_ci_contract.py` | `test_update_changelog_script_removed` |

## Regression Gate

>
> Accepted:
>   on non-zero exit.
>
> Rejected:

Workflow files:
Job names:
- `pr_checks_ruleset`
- `pr_checks_cc`
Specific failing commands each job runs:
- `pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py -q`
- `python3 tools/ruleset_gate.py --ruleset-file .github/rulesets/main.json --repo "$GITHUB_REPOSITORY" --ruleset-id 5406599`
- `python tools/cc_gate.py --pr-title "$PR_TITLE" --pr-body-file pr_body.txt --base-ref "$BASE" --head-ref "$HEAD" --repo "$GITHUB_REPOSITORY"`

Additionally, the repository ruleset for `refs/heads/main` must contain the 5-rule set `deletion`, `non_fast_forward`, `pull_request`, `copilot_code_review`, `required_status_checks`; the required-status-checks rule must keep `strict_required_status_checks_policy: true`; and the required contexts must be exactly `PR Checks CodeQL (python)`, `pr_checks_typing`, `pr_checks_cc`, `pr_checks_slice`, `pr_checks_fail_loud`, `pr_checks_version`, and `pr_checks_ruleset`.

## Out of Scope

- `tdw_control_plane/**`
- `Dockerfile*`
- `docker-compose*.yml`
- `.github/workflows/pr_checks_codeql.yml`
- `.github/workflows/pr_checks_slice_on_issue.yml`
- `tools/typing_gate.py`
- `tools/fail_loud_gate.py`
- `tools/version_gate.py`

## Risks and Open Questions

Risks:
- `pr_checks_tests` remains follow-up work after #157 lands; this slice stops at the 7-context ruleset state that can be rolled out immediately.
- This rollout uses an exclusive operator window. The operator captures the current live ruleset JSON, applies the target snapshot to ruleset `5406599` before opening the rollout PR, opens the PR from the branch that contains the matching snapshot and `pr_checks_ruleset` workflow, pauses other PR merges during the window, and immediately reapplies the captured pre-rollout JSON if the rollout is abandoned. `pr_checks_ruleset` has no bootstrap bypass; the live snapshot must already match before the rollout PR is expected to pass.
- `pr_checks_ruleset` reads repository ruleset `5406599` through `gh api` using the repo-hosted `pull_request` `GH_TOKEN`. On this public repository the call is available, but the returned payload can omit `bypass_actors`; if repo visibility or token policy changes and the call becomes unauthorized, the workflow must fail loudly rather than skip drift detection.
- `tools/ruleset_gate.py` now splits top-level live ruleset fields into REQUIRED `name`, `target`, `enforcement`, `conditions`, `rules` and OPTIONAL `bypass_actors`. The checked-in snapshot keeps both. The repo-hosted PR-path gate always compares the REQUIRED set, compares `bypass_actors` when the live payload exposes it, and logs `comparing observable subset only` when the token omits that OPTIONAL field. The stripped live API-only top-level fields remain `current_user_can_bypass`, `node_id`, `created_at`, `updated_at`, `_links`, `id`, `source`, and `source_type`; any other unexpected top-level live field is treated as drift until the snapshot and gate are deliberately updated together.
- Full live parity for `bypass_actors` moves to follow-up slice #163 because the repo-hosted PR-path token cannot mechanically prove that field today.
- Removing `pr_post_changelog.yml` is safe only because `pr_checks_version` already enforces pre-merge version/changelog updates on every PR.
- `pr_checks_ruleset` still runs as a repo-hosted `pull_request` workflow and therefore enforces repo-local drift detection on cooperative PRs, not immutable out-of-tree law. Hard-law anchoring outside the candidate tree remains follow-up work.

Open questions:
- None.

## Closeout

> bot verifies:

Merge SHA: (to be filled on merge)
Merged PR number: (to be filled on merge)

The closeout record must also include:
- the exact commands used to capture the pre-rollout live ruleset JSON, apply the target snapshot to ruleset `5406599` before opening the rollout PR, and roll back if needed
- command output showing the live ruleset on `refs/heads/main` still has the 5-rule set `deletion`, `non_fast_forward`, `pull_request`, `copilot_code_review`, `required_status_checks`
- command output showing `strict_required_status_checks_policy` remains `true`
- command output showing the required contexts are exactly `PR Checks CodeQL (python)`, `pr_checks_typing`, `pr_checks_cc`, `pr_checks_slice`, `pr_checks_fail_loud`, `pr_checks_version`, and `pr_checks_ruleset`

mikkokotila:
## Summary
- Add `.github/rulesets/main.json`, `tools/ruleset_gate.py`, `pr_checks_ruleset.yml`, and the fixture-backed ruleset/CI contract tests.
- Make `tools/cc_gate.py` fail loud when the linked-issue title lookup fails instead of silently skipping validation.
- Remove the post-merge changelog updater workflow and its script because `pr_checks_version` is the authoritative pre-merge gate.
- Keep `bypass_actors` in the checked-in ruleset snapshot, but make the repo-hosted PR-path gate compare only the live fields its token can actually observe and log that downgrade loudly.

## Testing
- `/tmp/tdw-cp-160-venv/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py -q`
- `python3 -m compileall tools tests`
- `python3 tools/version_gate.py --pr-title "ci: enforce required PR-path checks and ruleset drift on main" --base-pyproject <(git show origin/main:pyproject.toml) --head-pyproject pyproject.toml --base-changelog <(git show origin/main:CHANGELOG.md) --head-changelog CHANGELOG.md`
- `python3 tools/ruleset_gate.py --live-json tests/fixtures/github/ruleset_live_target.json --ruleset-file .github/rulesets/main.json --ruleset-id 5406599`
- `python3 tools/ruleset_gate.py --live-json tests/fixtures/github/ruleset_live_target_without_bypass_actors.json --ruleset-file .github/rulesets/main.json --ruleset-id 5406599`
- `python3 tools/ruleset_gate.py --live-json tests/fixtures/github/ruleset_live_with_bypass_actor.json --ruleset-file .github/rulesets/main.json --ruleset-id 5406599`
- `python3 tools/ruleset_gate.py --live-json tests/fixtures/github/ruleset_live_unexpected_field.json --ruleset-file .github/rulesets/main.json --ruleset-id 5406599`

## Notes
- This PR now targets the 7-context live ruleset state: `PR Checks CodeQL (python)`, `pr_checks_typing`, `pr_checks_cc`, `pr_checks_slice`, `pr_checks_fail_loud`, `pr_checks_version`, and `pr_checks_ruleset`.
- Ruleset `5406599` has already been updated to that 7-context target state on `main`.
- Follow-up [#162](https://github.com/Vaquum/tdw-control-plane/issues/162) widens the protected ruleset to include `pr_checks_tests` after [#157](https://github.com/Vaquum/tdw-control-plane/issues/157) lands.
- Follow-up [#163](https://github.com/Vaquum/tdw-control-plane/issues/163) adds a privileged full-parity ruleset audit for `bypass_actors`, because the repo-hosted `pull_request` token cannot mechanically prove that field today.
- Closes #160


tools/ruleset_gate.py
```diff
@@ -0,0 +1,163 @@
...
+    'name',
+    'target',
+    'enforcement',
+    'conditions',
+    'rules',
+})
+
+IGNORED_LIVE_FIELDS: Final[frozenset[str]] = frozenset({
+    '_links',
+    'bypass_actors',
```
mikkokotila: `bypass_actors` is ignored here, so the new drift gate stays green if someone weakens the live ruleset by adding a bypass actor. I reproduced that against this exact PR head by adding a synthetic `bypass_actors` entry to `tests/fixtures/github/ruleset_live_target.json` and running `python3 tools/ruleset_gate.py --live-json /tmp/ruleset_live_with_bypass_actor.json --ruleset-file .github/rulesets/main.json --ruleset-id 5406599`; it still returned `RULESET GATE -- PASS`. Exact fix: treat `bypass_actors` as part of the authoritative snapshot and fail on any drift there too. `current_user_can_bypass` can stay ignored because it is viewer-specific; `bypass_actors` cannot if this gate is meant to protect enforcement state.

mikkokotila (commented):
One code-level finding is left inline on `tools/ruleset_gate.py`: the gate currently ignores `bypass_actors`, so live ruleset weakening via a bypass grant is invisible to the drift check.

Separately, the current required blocker on this PR is `pr_checks_slice`: the live run failed because the PR title `ci: add ruleset drift gate on main` does not byte-equal issue #160 title `ci: enforce required PR-path checks and ruleset drift on main`.

mikkokotila:
## PR review against the issue #160 slice spec

The implementation tracks the slice spec closely. The workflow has the full plumbing the spec promised (checkout, setup-python 3.11, pip install pytest, `permissions: contents: read`, `GH_TOKEN` env on the gate step, `workflow_dispatch:` for re-runs). The fixtures are realistic captures with all API-only fields present. Tests are well-scoped — `test_issue_title_fetch_is_non_nullable_and_gate_has_no_none_guard` uses `source.index` to bound the type-signature assertions to the `fetch_issue_title` function block, then checks the dead-code guard across the whole file. Smart.

`tools/ruleset_gate.py` also includes the *"tiny sharpening"* I noted in the seventh-pass review — `normalize_live_ruleset` raises `SystemExit` with an explicit *"expected live ruleset field(s) missing"* message rather than letting the dict comprehension throw a bare `KeyError`. ([tools/ruleset_gate.py L115-122](https://github.com/Vaquum/tdw-control-plane/blob/db25f5d313ff0d19ab8a54b67b3899be0084359f/tools/ruleset_gate.py#L115-L122))

`tools/cc_gate.py::fetch_issue_title` adds one extra hardening beyond the slice spec: an empty title payload is now also `SystemExit(2)` ([cc_gate.py L174-179](https://github.com/Vaquum/tdw-control-plane/blob/db25f5d313ff0d19ab8a54b67b3899be0084359f/tools/cc_gate.py#L174-L179)). Defensive, consistent with the gate's contract, fine.

Two spec-vs-code drifts worth resolving before merge.

### Drift 1 — `IGNORED_LIVE_FIELDS` denylist has 9 entries in code, slice spec names 6

[tools/ruleset_gate.py L21-31](https://github.com/Vaquum/tdw-control-plane/blob/db25f5d313ff0d19ab8a54b67b3899be0084359f/tools/ruleset_gate.py#L21-L31):

```python
IGNORED_LIVE_FIELDS: Final[frozenset[str]] = frozenset({
    '_links', 'bypass_actors', 'created_at', 'current_user_can_bypass',
    'id', 'node_id', 'source', 'source_type', 'updated_at',
})
```

Three fields — `id`, `source`, `source_type` — are in the code but not in the issue body's Risks bullet, which says: *"tools/ruleset_gate.py strips only the named live API-only top-level fields `bypass_actors`, `current_user_can_bypass`, `node_id`, `created_at`, `updated_at`, and `_links`."*

The code is right. Those three fields are present in the actual live API response (verified earlier in this thread). If the code matched the slice text literally, every real API call would trip *"unexpected live ruleset field(s): ['id', 'source', 'source_type']"* and the gate would never pass. So the implementation correctly expanded the denylist; the slice text is now stale.

Two small fixes recommended:

- update the issue body's Risks bullet (and any prose that names the denylist) to list all 9 stripped fields, so a future maintainer reading the spec to validate the gate sees the same set the code sees
- consider adding a tiny `test_ignored_live_fields_match_named_set` test that asserts `IGNORED_LIVE_FIELDS == frozenset({...9 names...})` — pins the contract so a future PR cannot silently expand the denylist further (which would be the way to silently *miss* drift)

### Drift 2 — the PR touches `CHANGELOG.md` and `pyproject.toml`, but neither is in the slice's Surfaces

`pr_checks_version` requires every PR to modify both files (rules 1 and 3 in [version_gate.py](https://github.com/Vaquum/tdw-control-plane/blob/db25f5d313ff0d19ab8a54b67b3899be0084359f/tools/version_gate.py)). The slice's Surfaces section, by the slice template's own contract, must list every file the PR modifies. It currently lists only `tools/cc_gate.py` as Modified. `pr_checks_slice` is passing on this PR, which means `tools/slice_gate.py` either has a hardcoded exemption for those two files or its rule-7 enforcement is not as strict as the template prose implies.

Either way, the slice issue body should add `CHANGELOG.md` and `pyproject.toml` to the Modified list so the slice's mechanical scope contract matches what every PR genuinely touches under the version gate. Without it, every future slice PR will technically violate its own scope contract on these two files, even when the gate happens to allow it.

### Status note

`pr_checks_ruleset` is currently `failure`, which is the expected baseline state per the slice's rollout sequence — the live ruleset still has the 6-context state, snapshot has 8, drift exists by definition until the operator pre-applies the snapshot via `gh api --method PUT`. The PR's Notes section calls this out correctly. The actual rollout is also blocked on #157 landing first, since the snapshot requires `pr_checks_tests` as a context and that workflow doesn't exist on `main` yet. So the merge sequence is: #157 lands → operator captures pre-rollout live JSON → applies snapshot to ruleset 5406599 → re-runs `pr_checks_ruleset` on this PR via `workflow_dispatch` → all 8 contexts now report `success` → merge.

### Verdict

After the two drift items are resolved, this PR is ready for the rollout window. The implementation is faithful to the spec; the spec just needs to be updated to match a couple of details the implementation correctly discovered.

mikkokotila:
## Follow-up review — both drift findings resolved, plus three substantive new design changes

The two commits since my last review (`ad4d949` "fix ruleset drift review findings" and `9655036` "handle ruleset reads without bypass actors") resolve both items I flagged and add three substantive design improvements. All 8 required checks are now green on the head SHA, including `pr_checks_ruleset` — the operator has already pre-applied the new snapshot to live ruleset 5406599 (verified: live API returns 7 contexts + 5 rule types + `strict: true` + `bypass_actors: []`, matching the snapshot).

### What landed

**Drift 1 — denylist locked.** [`test_ignored_live_fields_match_named_set`](https://github.com/Vaquum/tdw-control-plane/blob/9655036b39bbb4ae2375d140ff80d994369d7254/tests/tools/test_ruleset_gate.py#L46-L60) now pins `IGNORED_LIVE_FIELDS` to a literal set of 8 names. Future PRs cannot silently expand the denylist.

**Drift 2 — pending in the slice body.** The PR still touches `CHANGELOG.md` and `pyproject.toml` (required by `version_gate`) without those paths appearing in issue #160's Surfaces. `pr_checks_slice` still passes, so the exemption is built into `slice_gate`, but the slice issue body should be updated to list these in Modified to keep the spec honest.

### Three substantive design changes worth recording

**1. `bypass_actors` promoted from IGNORED to SNAPSHOT.** Previously `bypass_actors` was stripped before comparison. Now it is part of the compared snapshot, with `bypass_actors: []` baked into [.github/rulesets/main.json L13](https://github.com/Vaquum/tdw-control-plane/blob/9655036b39bbb4ae2375d140ff80d994369d7254/.github/rulesets/main.json#L13) and a [`test_bypass_actor_drift_is_failure`](https://github.com/Vaquum/tdw-control-plane/blob/9655036b39bbb4ae2375d140ff80d994369d7254/tests/tools/test_ruleset_gate.py#L39-L43) test pinning the failure mode. This is a real security improvement — adding a bypass actor to the live ruleset is now drift the gate catches. Was not in my review; spotted and added on your initiative.

**2. `pr_checks_tests` dropped from required contexts.** The snapshot now has 7 contexts (no `pr_checks_tests`) and the live API has been mutated to match. This decouples this slice from #157 — the rollout no longer needs to wait for #157 to land. That is a substantive descope; the slice spec (issue #160) still says 8 contexts including `pr_checks_tests` in Target State and Closeout, and the Risks bullet still says *"until #157 lands, the live target state for this slice cannot be completed."* Reconcile in the issue body — either drop `pr_checks_tests` from the Target/Closeout (matches reality, this slice no longer blocks on #157) or restore it to the snapshot (and re-block on #157). The operational reality is the former.

**3. `bypass_actors` is OPTIONAL on live, with `comparing observable subset only` warning when absent.** The split into [`REQUIRED_TOP_LEVEL_FIELDS`](https://github.com/Vaquum/tdw-control-plane/blob/9655036b39bbb4ae2375d140ff80d994369d7254/tools/ruleset_gate.py#L13-L19) and [`OPTIONAL_LIVE_TOP_LEVEL_FIELDS`](https://github.com/Vaquum/tdw-control-plane/blob/9655036b39bbb4ae2375d140ff80d994369d7254/tools/ruleset_gate.py#L21-L23) handles the case where the live API omits `bypass_actors` for non-admin-scoped reads. Elegant solution to a real operational constraint. **But** — this introduces a narrow fail-silent surface: in any CI run where the token cannot read `bypass_actors`, bypass-actor drift is silently accepted. The stderr warning is loud-ish but does not fail the gate.

This is the right operational trade-off (the alternative — gate always fails on PR-time runs without admin scope — would make the gate unusable). It is consistent with the principle "*govern everything they have capability for handling*" because the gate explicitly announces when its capability is reduced.

But two completions would close the loop:

- name this trade-off in the slice spec — currently the Risks bullet implies the gate strips a fixed denylist and treats everything else as drift; the actual behavior is more nuanced (REQUIRED always compared, OPTIONAL compared when present, IGNORED always stripped)
- add a sibling slice for a **scheduled admin-scoped drift probe** — a workflow that runs on `schedule:` with a PAT that has `administration: read`, calls the same gate against the live ruleset (where `bypass_actors` IS visible), and fails loud if drift exists. That recovers the bypass-actor enforcement that PR-time gates cannot guarantee.

### Smaller note

There is no equivalent set-membership pin test for `REQUIRED_TOP_LEVEL_FIELDS` or `OPTIONAL_LIVE_TOP_LEVEL_FIELDS` — only for `IGNORED_LIVE_FIELDS`. A future PR that adds a field to `REQUIRED` (potentially a security-relevant one) or moves a field from REQUIRED to OPTIONAL (silently weakening the comparison) would not be caught. Mirroring the existing pin test for the other two sets closes that gap. One-liner each.

### CI status

All 8 required checks green on head `9655036`. Live ruleset matches snapshot. PR's `mergeable_state: blocked` is from the `pull_request` rule's `required_review_thread_resolution` or Copilot review, not from any failing check.

### Verdict

The implementation is in good shape. Three things to land before merge:

1. update issue #160 body to list `CHANGELOG.md` and `pyproject.toml` in Modified
2. update issue #160 body to drop `pr_checks_tests` from Target/Closeout (and remove the #157 dependency in Risks) — matches operational reality
3. update issue #160 Risks bullet that names the denylist to also describe the REQUIRED/OPTIONAL split and the "comparing observable subset only" behavior

Plus the two-line set-membership pin tests for `REQUIRED_TOP_LEVEL_FIELDS` and `OPTIONAL_LIVE_TOP_LEVEL_FIELDS` if you want symmetric protection.

Sibling-slice candidate (not blocking): scheduled admin-scoped drift probe for `bypass_actors` enforcement.

PR #165, opened 2026-04-22
ci: add pr_checks_lint for tools and tests/tools on main

### LINKED ISSUE #164: ci: add pr_checks_lint for tools and tests/tools on main
## Slice ID

S05

## Slice Title

ci: add pr_checks_lint for tools and tests/tools on main

## Routing

Parent issue: none
Depends on:
- #160
Unlocks:
- Protected PR-path Ruff enforcement for the repo's CI/tooling Python surfaces on `main`
- A required `pr_checks_lint` context in the checked-in and live `main` ruleset snapshot
- Future expansion from CI/tooling-only lint enforcement to wider repo surfaces after the current 206-error full-repo Ruff baseline is paid down

## Minimum Viable Capability

>
> Accepted:
>
> Rejected:
> - *Strict CQRS separation.*

- `test -f .github/workflows/pr_checks_lint.yml; echo $?` returns `0`
- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context | @json'` returns the 8-line set `"PR Checks CodeQL (python)"`, `"pr_checks_typing"`, `"pr_checks_cc"`, `"pr_checks_slice"`, `"pr_checks_fail_loud"`, `"pr_checks_version"`, `"pr_checks_ruleset"`, `"pr_checks_lint"`
- `python3 -m venv /tmp/tdw-cp-lint-venv && /tmp/tdw-cp-lint-venv/bin/python -m pip install --upgrade pip 'ruff==0.15.11' >/dev/null && /tmp/tdw-cp-lint-venv/bin/python -m ruff check tools tests/tools` exits `0`
- `python3 -m venv /tmp/tdw-cp-lint-venv && /tmp/tdw-cp-lint-venv/bin/python -m pip install --upgrade pip 'ruff==0.15.11' >/dev/null && /tmp/tdw-cp-lint-venv/bin/python -m ruff check tests/fixtures/lint/bad_imports.py` exits `1`
- `python3 -c "from pathlib import Path; import re; files=['.github/workflows/pr_checks_lint.yml','.github/workflows/pr_checks_ruleset.yml']; pins=sorted({m for f in files for m in re.findall(r'ruff==([0-9.]+)', Path(f).read_text())}); print(pins)"` returns `['0.15.11']`
- `python3 -c "import json,pathlib,tomllib; data = tomllib.loads(pathlib.Path('pyproject.toml').read_text()); r = data['tool']['ruff']; print(json.dumps({'exclude': r.get('exclude'), 'ignore': r['lint'].get('ignore'), 'per_file_ignores': r['lint'].get('per-file-ignores'), 'select': r['lint'].get('select')}, sort_keys=True))"` returns `{"exclude": [".git", "__pycache__", "build", "dist", "quickstart_etl_tests"], "ignore": ["E501"], "per_file_ignores": {"quickstart_etl_tests/**/*.py": ["S101", "ANN", "BLE001"]}, "select": ["E", "F", "I", "UP", "RUF", "BLE", "ANN"]}`

## Current State

>
> Accepted:
>
> Rejected:

- `test -f .github/workflows/pr_checks_lint.yml; echo $?` returns `1`
- `gh workflow list --repo Vaquum/tdw-control-plane | rg '^pr_checks_lint$'` exits `1`
- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context | @json'` returns the 7-line set `"PR Checks CodeQL (python)"`, `"pr_checks_typing"`, `"pr_checks_cc"`, `"pr_checks_slice"`, `"pr_checks_fail_loud"`, `"pr_checks_version"`, `"pr_checks_ruleset"`
- `uvx ruff check tools tests/tools --statistics 2>&1 | rg '^(warning: The following rules have been removed and ignoring them has no effect:|    - ANN101|    - ANN102|[0-9]+\t[A-Z0-9]+|Found [0-9]+ errors\.$)'` exits `0` and returns the 7-line set `warning: The following rules have been removed and ignoring them has no effect:`, `    - ANN101`, `    - ANN102`, `3	I001  	[*] unsorted-imports`, `2	BLE001	[ ] blind-except`, `2	RUF005	[ ] collection-literal-concatenation`, `Found 7 errors.`
- `uvx ruff check . 2>&1 | rg '^Found [0-9]+ errors\.$' | head -n 1` returns `Found 206 errors.`
- `python3 -c "import json,pathlib,tomllib; data = tomllib.loads(pathlib.Path('pyproject.toml').read_text()); r = data['tool']['ruff']; print(json.dumps({'exclude': r.get('exclude'), 'ignore': r['lint'].get('ignore'), 'per_file_ignores': r['lint'].get('per-file-ignores'), 'select': r['lint'].get('select')}, sort_keys=True))"` returns `{"exclude": [".git", "__pycache__", "build", "dist", "quickstart_etl_tests"], "ignore": ["E501", "ANN101", "ANN102"], "per_file_ignores": {"quickstart_etl_tests/**/*.py": ["S101", "ANN", "BLE001"]}, "select": ["E", "F", "I", "UP", "RUF", "BLE", "ANN"]}`

## Target State

> predicate.
>
> Accepted:

- `test -f .github/workflows/pr_checks_lint.yml; echo $?` returns `0`
- `gh workflow list --repo Vaquum/tdw-control-plane | rg '^pr_checks_lint$'` returns one active workflow line
- `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599 --jq '.rules[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context | @json'` returns the 8-line set `"PR Checks CodeQL (python)"`, `"pr_checks_typing"`, `"pr_checks_cc"`, `"pr_checks_slice"`, `"pr_checks_fail_loud"`, `"pr_checks_version"`, `"pr_checks_ruleset"`, `"pr_checks_lint"`
- `python3 -m venv /tmp/tdw-cp-lint-venv && /tmp/tdw-cp-lint-venv/bin/python -m pip install --upgrade pip 'ruff==0.15.11' >/dev/null && /tmp/tdw-cp-lint-venv/bin/python -m ruff check tools tests/tools` exits `0`
- `python3 -m venv /tmp/tdw-cp-lint-venv && /tmp/tdw-cp-lint-venv/bin/python -m pip install --upgrade pip 'ruff==0.15.11' >/dev/null && /tmp/tdw-cp-lint-venv/bin/python -m ruff check tests/fixtures/lint/bad_imports.py` exits `1`
- `python3 -c "from pathlib import Path; import re; files=['.github/workflows/pr_checks_lint.yml','.github/workflows/pr_checks_ruleset.yml']; pins=sorted({m for f in files for m in re.findall(r'ruff==([0-9.]+)', Path(f).read_text())}); print(pins)"` returns `['0.15.11']`
- `python3 -c "import json,pathlib,tomllib; data = tomllib.loads(pathlib.Path('pyproject.toml').read_text()); r = data['tool']['ruff']; print(json.dumps({'exclude': r.get('exclude'), 'ignore': r['lint'].get('ignore'), 'per_file_ignores': r['lint'].get('per-file-ignores'), 'select': r['lint'].get('select')}, sort_keys=True))"` returns `{"exclude": [".git", "__pycache__", "build", "dist", "quickstart_etl_tests"], "ignore": ["E501"], "per_file_ignores": {"quickstart_etl_tests/**/*.py": ["S101", "ANN", "BLE001"]}, "select": ["E", "F", "I", "UP", "RUF", "BLE", "ANN"]}`

## Surfaces

- `CHANGELOG.md`
- `pyproject.toml`
- `tools/fail_loud_gate.py`
- `tools/slice_gate.py`
- `tools/typing_gate.py`
- `tests/tools/test_cc_gate.py`
- `tests/tools/test_ci_contract.py`

Added:
- `.github/workflows/pr_checks_lint.yml`
- `tests/tools/test_lint_ci_contract.py`
- `tests/fixtures/lint/bad_imports.py`

Removed:
- none

## Signature Diff

### .github/workflows/pr_checks_lint.yml::jobs.pr_checks_lint
Before:
```
(file does not exist)
```
After:
```yaml
name: pr_checks_lint

on:
  pull_request:
    branches: [main]
  workflow_dispatch:

permissions:
  contents: read

jobs:
  pr_checks_lint:
    name: pr_checks_lint
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v4
      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version: '3.11'
      - name: Set up lint venv
        run: python -m venv .venv-lint
      - name: Install Ruff
        run: .venv-lint/bin/python -m pip install --upgrade pip 'ruff==0.15.11'
      - name: Run Ruff lint gate
        run: .venv-lint/bin/python -m ruff check tools tests/tools
```

### .github/workflows/pr_checks_ruleset.yml::jobs.pr_checks_ruleset.steps
Before:
```yaml
      - name: Install test dependency
        run: python -m pip install --upgrade pip pytest

      - name: Run CI contract tests
        run: pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py -q
```
After:
```yaml
      - name: Set up contract venv
        run: python -m venv .venv-ruleset

      - name: Install contract dependencies
        run: .venv-ruleset/bin/python -m pip install --upgrade pip pytest 'ruff==0.15.11'

      - name: Run CI contract tests
        run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py -q
```

### .github/rulesets/main.json::required_status_checks
Before:
```json
[
  "PR Checks CodeQL (python)",
  "pr_checks_typing",
  "pr_checks_cc",
  "pr_checks_slice",
  "pr_checks_fail_loud",
  "pr_checks_version",
  "pr_checks_ruleset"
]
```
After:
```json
[
  "PR Checks CodeQL (python)",
  "pr_checks_typing",
  "pr_checks_cc",
  "pr_checks_slice",
  "pr_checks_fail_loud",
  "pr_checks_version",
  "pr_checks_ruleset",
  "pr_checks_lint"
]
```

### pyproject.toml::tool.ruff.exclude + lint.select + lint.ignore + per-file-ignores
Before:
```toml
exclude = [
  ".git",
  "__pycache__",
  "build",
  "dist",
  "quickstart_etl_tests",
]

select = [
  "E",     # pycodestyle errors
  "F",     # pyflakes
  "I",     # isort
  "UP",    # pyupgrade
  "RUF",   # ruff-specific
  "BLE",   # flake8-blind-except -- always raise, never bare except
  "ANN",   # flake8-annotations -- catches missing type annotations
]

ignore = [
  "E501",   # line length handled by formatter
  "ANN101", # self typing not needed
  "ANN102", # cls typing not needed
]

[tool.ruff.lint.per-file-ignores]
"quickstart_etl_tests/**/*.py" = ["S101", "ANN", "BLE001"]
```
After:
```toml
exclude = [
  ".git",
  "__pycache__",
  "build",
  "dist",
  "quickstart_etl_tests",
]

select = [
  "E",     # pycodestyle errors
  "F",     # pyflakes
  "I",     # isort
  "UP",    # pyupgrade
  "RUF",   # ruff-specific
  "BLE",   # flake8-blind-except -- always raise, never bare except
  "ANN",   # flake8-annotations -- catches missing type annotations
]

ignore = [
  "E501",   # line length handled by formatter
]

[tool.ruff.lint.per-file-ignores]
"quickstart_etl_tests/**/*.py" = ["S101", "ANN", "BLE001"]
```

### tools/fail_loud_gate.py::gate
Before:
```python
    root_name = budget_head.get('package_root')
    if not isinstance(root_name, str) or not root_name:
        return failures + ['fail_loud_gate: budget must set `package_root`']
    package_root = REPO_ROOT / root_name
    if not package_root.is_dir():
        return failures + [
            f'fail_loud_gate: package_root {root_name!r} not found under repo root'
        ]
```
After:
```python
    root_name = budget_head.get('package_root')
    if not isinstance(root_name, str) or not root_name:
        return [*failures, 'fail_loud_gate: budget must set `package_root`']
    package_root = REPO_ROOT / root_name
    if not package_root.is_dir():
        return [
            *failures,
            f'fail_loud_gate: package_root {root_name!r} not found under repo root',
        ]
```

### tools/slice_gate.py::read_lines
Before:
```python
def read_lines(path: Path) -> list[str]:
    try:
        text = path.read_text(encoding='utf-8')
    except OSError as exc:
        raise SystemExit(
            f'slice_gate: cannot read {path}: {exc}'
        ) from exc
    return [line for line in text.splitlines() if line.strip()]
```
After:
```python
def read_lines(path: Path) -> list[str]:
    try:
        text = path.read_text(encoding='utf-8')
    except OSError as exc:
        print(
            f'slice_gate: cannot read {path}: {exc}',
            file=sys.stderr,
        )
        raise SystemExit(2) from exc
    return [line for line in text.splitlines() if line.strip()]
```

### tools/typing_gate.py::_setup_failure + load_pyproject + load_budget
Before:
```python
def load_pyproject() -> dict[str, object]:
    with open(PYPROJECT_PATH, 'rb') as f:
        return tomllib.load(f)


def load_budget() -> dict[str, object]:
    if not BUDGET_PATH.exists():
        raise SystemExit(
            f'typing gate cannot run: {BUDGET_PATH.relative_to(REPO_ROOT)} '
            f'is missing. Run with --update-budget to create it.'
        )
    return json.loads(BUDGET_PATH.read_text())
```
After:
```python
def _setup_failure(message: str) -> None:
    print(message, file=sys.stderr)
    raise SystemExit(2)


def load_pyproject() -> dict[str, object]:
    try:
        with open(PYPROJECT_PATH, 'rb') as f:
            return tomllib.load(f)
    except OSError as exc:
        _setup_failure(
            f'typing_gate: cannot read {PYPROJECT_PATH.relative_to(REPO_ROOT)}: {exc}'
        )
    except tomllib.TOMLDecodeError as exc:
        _setup_failure(
            f'typing_gate: cannot parse {PYPROJECT_PATH.relative_to(REPO_ROOT)}: {exc}'
        )


def load_budget() -> dict[str, object]:
    if not BUDGET_PATH.exists():
        _setup_failure(
            f'typing_gate: cannot run: {BUDGET_PATH.relative_to(REPO_ROOT)} '
            f'is missing. Run with --update-budget to create it.'
        )
    try:
        return json.loads(BUDGET_PATH.read_text(encoding='utf-8'))
    except OSError as exc:
        _setup_failure(
            f'typing_gate: cannot read {BUDGET_PATH.relative_to(REPO_ROOT)}: {exc}'
        )
    except json.JSONDecodeError as exc:
        _setup_failure(
            f'typing_gate: cannot parse {BUDGET_PATH.relative_to(REPO_ROOT)}: {exc}'
        )
```

### tests/fixtures/lint/bad_imports.py::header
Before:
```
(file does not exist)
```
After:
```python
# Intentionally lint-broken. Do not "fix." Used by
# tests/tools/test_lint_ci_contract.py::test_pinned_ruff_fails_on_known_bad_fixture.
import os, sys
```

## Tests That Prove the MVC

|---|---|---|
| lint workflow exists | `tests/tools/test_lint_ci_contract.py` | `test_pr_checks_lint_workflow_exists` |
| ruleset snapshot requires `pr_checks_lint` | `tests/tools/test_lint_ci_contract.py` | `test_ruleset_snapshot_requires_pr_checks_lint` |
| pinned Ruff gate targets only `tools` and `tests/tools` | `tests/tools/test_lint_ci_contract.py` | `test_pr_checks_lint_runs_pinned_ruff_on_tools_and_tests_tools` |
| required CI contract runner executes the lint contract test | `tests/tools/test_lint_ci_contract.py` | `test_pr_checks_ruleset_runs_test_lint_ci_contract` |
| a known bad Ruff fixture fails under the pinned invocation | `tests/tools/test_lint_ci_contract.py` | `test_pinned_ruff_fails_on_known_bad_fixture` |
| Ruff pin is consistent across the lint gate and contract runner workflows | `tests/tools/test_lint_ci_contract.py` | `test_ruff_pin_is_consistent_across_workflows` |
| authoritative Ruff policy surface (`exclude`, `select`, `ignore`, `per-file-ignores`) stays exact and isolated `BLE001` stays clean across `tools tests/tools` | `tests/tools/test_lint_ci_contract.py` | `test_pyproject_ruff_policy_contract` |

## Regression Gate

>
> Accepted:
>   on non-zero exit.
>
> Rejected:

Workflow files:
- `.github/workflows/pr_checks_lint.yml`
Job names:
- `pr_checks_lint`
- `pr_checks_ruleset`
Specific failing commands each job runs:
- `.venv-lint/bin/python -m ruff check tools tests/tools`
- `.venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py -q`

## Out of Scope

- `tdw_control_plane/**`
- `.github/workflows/pr_checks_codeql.yml`
- `tools/version_gate.py`

## Risks and Open Questions

Risks:
- Full-repo Ruff enforcement is explicitly out of scope for this slice because `uvx ruff check .` currently reports `206` errors. This slice stops at full fail-hard enforcement of the CI/tooling path `tools` plus `tests/tools`.
- The current scoped baseline is exactly `3 I001`, `2 BLE001`, and `2 RUF005`. This slice fixes `I001`, `BLE001`, and `RUF005` directly. No new Ruff ignores or per-file exemptions are introduced; the two broad-exception sites in `tools/slice_gate.py` and `tools/typing_gate.py` are replaced with explicit fail-loud read/setup handling. `tests/tools/test_lint_ci_contract.py::test_pyproject_ruff_policy_contract` pins the authoritative Ruff policy surface (`exclude`, `select`, `ignore`, `per-file-ignores`) and runs `ruff check --isolated --select BLE001 tools tests/tools` so a candidate PR cannot relax its own lint law and keep the contract green.
- `tests/tools/test_cc_gate.py` and `tests/tools/test_ruleset_gate.py` are in scope only for `I001` import-order normalization. `tests/tools/test_ci_contract.py` also carries a regression test proving `tools/typing_gate.py` still prints a `typing_gate:`-prefixed setup error to stderr and exits `2` when `pyproject.toml` is unreadable or unparsable.
- The rollout must inherit the `#160` operator-window sequence because `pr_checks_ruleset` is already required on `main`: capture the current live ruleset JSON, apply the 8-context target snapshot to ruleset `5406599` before opening or expecting green on the rollout PR, pause merges during the window, and reapply the captured pre-rollout JSON if the rollout is abandoned.
- `pr_checks_lint` and the `pr_checks_ruleset` contract runner must install pinned Ruff `0.15.11` and fail loud on install or execution error; no warning-only path, `continue-on-error`, or best-effort skip is in contract.
- Ruff pin maintenance is follow-up work. Bumping beyond `0.15.11` is its own slice because new Ruff releases can change rule behavior and expose fresh findings in the gated scope.
- Ruff `0.15.11` is hardcoded in both `.github/workflows/pr_checks_lint.yml` and `.github/workflows/pr_checks_ruleset.yml`; `tests/tools/test_lint_ci_contract.py::test_ruff_pin_is_consistent_across_workflows` must pin those two workflow surfaces to the same literal version so the gate and its contract runner cannot silently drift apart.

Open questions:
- None.

## Closeout

> bot verifies:

Merge SHA:
Merged PR number:
- `pr_checks_lint`: (to be filled on merge)

The closeout record must also include:
- the exact commands used to capture the pre-rollout live ruleset JSON, apply the 8-context target snapshot to ruleset `5406599` before opening the rollout PR, and roll back if needed
- command output showing the live ruleset on `refs/heads/main` still has the 5-rule set `deletion`, `non_fast_forward`, `pull_request`, `copilot_code_review`, `required_status_checks`
- command output showing `strict_required_status_checks_policy` remains `true`
- command output showing the required contexts are exactly `PR Checks CodeQL (python)`, `pr_checks_typing`, `pr_checks_cc`, `pr_checks_slice`, `pr_checks_fail_loud`, `pr_checks_version`, `pr_checks_ruleset`, and `pr_checks_lint`

## Author Checks

mikkokotila:
Closes #164

## What changed
- add `.github/workflows/pr_checks_lint.yml` to run Ruff `0.15.11` against `tools` and `tests/tools`
- extend `.github/workflows/pr_checks_ruleset.yml` so required CI also runs `tests/tools/test_lint_ci_contract.py` with the same pinned Ruff version
- update `.github/rulesets/main.json` and the ruleset fixtures to require `pr_checks_lint`
- remove the dead Ruff ignores `ANN101` and `ANN102`, add exact `BLE001` per-file ignores for `tools/slice_gate.py` and `tools/typing_gate.py`, and replace the remaining `RUF005` list concatenations in `tools/fail_loud_gate.py`
- add the lint contract test module and intentionally broken Ruff fixture for negative-path proof
- bump the package version to `1.3.3` and record the slice in `CHANGELOG.md`
- apply the live `Protect-Main` ruleset update to add `pr_checks_lint` before opening this rollout PR, per the `#164` operator-window sequence

## Local verification
- `python3 -m venv /tmp/tdw-cp-lint-venv && /tmp/tdw-cp-lint-venv/bin/python -m pip install --upgrade pip pytest 'ruff==0.15.11' >/dev/null && /tmp/tdw-cp-lint-venv/bin/python -m ruff check tools tests/tools`
- `python3 -m venv /tmp/tdw-cp-lint-venv && /tmp/tdw-cp-lint-venv/bin/python -m pip install --upgrade pip pytest 'ruff==0.15.11' >/dev/null && /tmp/tdw-cp-lint-venv/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py -q`
- `python3 -m compileall tools tests`
- `python3 tools/version_gate.py --pr-title 'ci: add pr_checks_lint for tools and tests/tools on main' --base-pyproject <(git show origin/main:pyproject.toml) --head-pyproject pyproject.toml --base-changelog <(git show origin/main:CHANGELOG.md) --head-changelog CHANGELOG.md`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,8 +109,6 @@ ignore = [
 
 [tool.ruff.lint.per-file-ignores]
 "quickstart_etl_tests/**/*.py" = ["S101", "ANN", "BLE001"]
-"tools/slice_gate.py" = ["BLE001"]
-"tools/typing_gate.py" = ["BLE001"]
 
 [tool.ruff.lint.isort]
 known-first-party = ["tdw_control_plane"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "quickstart_etl"
-version = "1.3.2"
+version = "1.3.3"
 description = "Trades Warehouse Control Plane"
 readme = "README.md"
 requires-python = ">=3.11"
@@ -105,12 +105,12 @@ select = [
 ]
 ignore = [
   "E501",   # line length handled by formatter
-  "ANN101", # self typing not needed
-  "ANN102", # cls typing not needed
 ]
 
 [tool.ruff.lint.per-file-ignores]
 "quickstart_etl_tests/**/*.py" = ["S101", "ANN", "BLE001"]
+"tools/slice_gate.py" = ["BLE001"]
+"tools/typing_gate.py" = ["BLE001"]
 
 [tool.ruff.lint.isort]
 known-first-party = ["tdw_control_plane"]
--- a/tests/tools/test_lint_ci_contract.py
+++ b/tests/tools/test_lint_ci_contract.py
@@ -1,0 +1,81 @@
+from __future__ import annotations
+
+import json
+import re
+import subprocess
+import sys
+import tomllib
+from pathlib import Path
+from typing import Final
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+LINT_WORKFLOW: Final[Path] = REPO_ROOT / '.github/workflows/pr_checks_lint.yml'
+RULESET_WORKFLOW: Final[Path] = REPO_ROOT / '.github/workflows/pr_checks_ruleset.yml'
+RULESET_SNAPSHOT: Final[Path] = REPO_ROOT / '.github/rulesets/main.json'
+BAD_FIXTURE: Final[Path] = REPO_ROOT / 'tests/fixtures/lint/bad_imports.py'
+RUFF_VERSION: Final[str] = '0.15.11'
+
+
+def _required_status_contexts() -> list[str]:
+    payload = json.loads(RULESET_SNAPSHOT.read_text(encoding="utf-8"))
+    for rule in payload['rules']:
+        if rule['type'] == 'required_status_checks':
+            checks = rule['parameters']['required_status_checks']
+            return [entry['context'] for entry in checks]
+    raise AssertionError('required_status_checks rule missing from ruleset snapshot')
+
+
+def test_pr_checks_lint_workflow_exists() -> None:
+    assert LINT_WORKFLOW.exists()
+
+
+def test_ruleset_snapshot_requires_pr_checks_lint() -> None:
+    assert 'pr_checks_lint' in _required_status_contexts()
+
+
+def test_pr_checks_lint_runs_pinned_ruff_on_tools_and_tests_tools() -> None:
+    workflow = LINT_WORKFLOW.read_text(encoding='utf-8')
+
+    assert f"'ruff=={RUFF_VERSION}'" in workflow
+    assert '.venv-lint/bin/python -m ruff check tools tests/tools' in workflow
+    assert 'continue-on-error' not in workflow
+
+
+def test_pr_checks_ruleset_runs_test_lint_ci_contract() -> None:
+    workflow = RULESET_WORKFLOW.read_text(encoding='utf-8')
+
+    assert f"'ruff=={RUFF_VERSION}'" in workflow
+    assert 'tests/tools/test_lint_ci_contract.py' in workflow
+
+
+def test_pinned_ruff_fails_on_known_bad_fixture() -> None:
+    result = subprocess.run(
+        [sys.executable, '-m', 'ruff', 'check', str(BAD_FIXTURE)],
+        check=False,
+        capture_output=True,
+        text=True,
+        cwd=REPO_ROOT,
+    )
+
+    assert result.returncode == 1
+    assert 'bad_imports.py' in f'{result.stdout}\n{result.stderr}'
+
+
+def test_ruff_pin_is_consistent_across_workflows() -> None:
+    files = [LINT_WORKFLOW, RULESET_WORKFLOW]
+    pins = sorted({
+        pin
+        for workflow in files
+        for pin in re.findall(r'ruff==([0-9.]+)', workflow.read_text(encoding='utf-8'))
+    })
+
+    assert pins == [RUFF_VERSION]
+
+
+def test_pyproject_ruff_ignore_contract() -> None:
+    data = tomllib.loads((REPO_ROOT / 'pyproject.toml').read_text(encoding='utf-8'))
+    lint = data['tool']['ruff']['lint']
+
+    assert sorted(lint['ignore']) == ['E501']
+    assert lint['per-file-ignores']['tools/slice_gate.py'] == ['BLE001']
+    assert lint['per-file-ignores']['tools/typing_gate.py'] == ['BLE001']
--- a/tests/tools/test_lint_ci_contract.py
+++ b/tests/tools/test_lint_ci_contract.py
@@ -14,6 +14,28 @@ RULESET_WORKFLOW: Final[Path] = REPO_ROOT / '.github/workflows/pr_checks_ruleset
 RULESET_SNAPSHOT: Final[Path] = REPO_ROOT / '.github/rulesets/main.json'
 BAD_FIXTURE: Final[Path] = REPO_ROOT / 'tests/fixtures/lint/bad_imports.py'
 RUFF_VERSION: Final[str] = '0.15.11'
+EXPECTED_RUFF_POLICY: Final[dict[str, object]] = {
+    'exclude': [
+        '.git',
+        '__pycache__',
+        'build',
+        'dist',
+        'quickstart_etl_tests',
+    ],
+    'select': [
+        'E',
+        'F',
+        'I',
+        'UP',
+        'RUF',
+        'BLE',
+        'ANN',
+    ],
+    'ignore': ['E501'],
+    'per-file-ignores': {
+        'quickstart_etl_tests/**/*.py': ['S101', 'ANN', 'BLE001'],
+    },
+}
 
 
 def _required_status_contexts() -> list[str]:
@@ -80,14 +102,17 @@ def test_ruff_pin_is_consistent_across_workflows() -> None:
     assert pins == [RUFF_VERSION]
 
 
-def test_pyproject_ruff_ignore_contract() -> None:
+def test_pyproject_ruff_policy_contract() -> None:
     data = tomllib.loads((REPO_ROOT / 'pyproject.toml').read_text(encoding='utf-8'))
-    lint = data['tool']['ruff']['lint']
+    ruff = data['tool']['ruff']
+    actual_policy = {
+        'exclude': ruff.get('exclude'),
+        'select': ruff['lint'].get('select'),
+        'ignore': ruff['lint'].get('ignore'),
+        'per-file-ignores': ruff['lint'].get('per-file-ignores'),
+    }
 
-    assert sorted(lint['ignore']) == ['E501']
-    assert 'tools/slice_gate.py' not in lint['per-file-ignores']
-    assert 'tools/typing_gate.py' not in lint['per-file-ignores']
+    assert actual_policy == EXPECTED_RUFF_POLICY
 
-    for path in ('tools/slice_gate.py', 'tools/typing_gate.py'):
-        result = _run_ruff('check', '--isolated', '--select', 'BLE001', path)
-        assert result.returncode == 0, result.stdout + result.stderr
+    result = _run_ruff('check', '--isolated', '--select', 'BLE001', 'tools', 'tests/tools')
+    assert result.returncode == 0, result.stdout + result.stderr
--- a/tests/tools/test_lint_ci_contract.py
+++ b/tests/tools/test_lint_ci_contract.py
@@ -17,12 +17,22 @@ RUFF_VERSION: Final[str] = '0.15.11'
 
 
 def _required_status_contexts() -> list[str]:
-    payload = json.loads(RULESET_SNAPSHOT.read_text(encoding="utf-8"))
+    payload = json.loads(RULESET_SNAPSHOT.read_text(encoding='utf-8'))
     for rule in payload['rules']:
         if rule['type'] == 'required_status_checks':
             checks = rule['parameters']['required_status_checks']
             return [entry['context'] for entry in checks]
     raise AssertionError('required_status_checks rule missing from ruleset snapshot')
+
+
+def _run_ruff(*args: str) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(
+        [sys.executable, '-m', 'ruff', *args],
+        check=False,
+        capture_output=True,
+        text=True,
+        cwd=REPO_ROOT,
+    )
 
 
 def test_pr_checks_lint_workflow_exists() -> None:
@@ -49,13 +59,11 @@ def test_pr_checks_ruleset_runs_test_lint_ci_contract() -> None:
 
 
 def test_pinned_ruff_fails_on_known_bad_fixture() -> None:
-    result = subprocess.run(
-        [sys.executable, '-m', 'ruff', 'check', str(BAD_FIXTURE)],
-        check=False,
-        capture_output=True,
-        text=True,
-        cwd=REPO_ROOT,
-    )
+    version = _run_ruff('--version')
+    assert version.returncode == 0, version.stderr
+    assert version.stdout.strip() == f'ruff {RUFF_VERSION}'
+
+    result = _run_ruff('check', str(BAD_FIXTURE))
 
     assert result.returncode == 1
     assert 'bad_imports.py' in f'{result.stdout}\n{result.stderr}'
@@ -77,5 +85,9 @@ def test_pyproject_ruff_ignore_contract() -> None:
     lint = data['tool']['ruff']['lint']
 
     assert sorted(lint['ignore']) == ['E501']
-    assert lint['per-file-ignores']['tools/slice_gate.py'] == ['BLE001']
-    assert lint['per-file-ignores']['tools/typing_gate.py'] == ['BLE001']
+    assert 'tools/slice_gate.py' not in lint['per-file-ignores']
+    assert 'tools/typing_gate.py' not in lint['per-file-ignores']
+
+    for path in ('tools/slice_gate.py', 'tools/typing_gate.py'):
+        result = _run_ruff('check', '--isolated', '--select', 'BLE001', path)
+        assert result.returncode == 0, result.stdout + result.stderr
--- a/tools/typing_gate.py
+++ b/tools/typing_gate.py
@@ -56,6 +56,11 @@ BUDGET_PATH: Final[Path] = REPO_ROOT / '.github' / 'typing_budget.json'
 PYPROJECT_PATH: Final[Path] = REPO_ROOT / 'pyproject.toml'
 
 
+def _setup_failure(message: str) -> None:
+    print(message, file=sys.stderr)
+    raise SystemExit(2)
+
+
 # -------------------------------------------------------------------
 # File walking
 # -------------------------------------------------------------------
@@ -867,31 +872,31 @@ def load_pyproject() -> dict[str, object]:
         with open(PYPROJECT_PATH, 'rb') as f:
             return tomllib.load(f)
     except OSError as exc:
-        raise SystemExit(
-            f'typing gate cannot read {PYPROJECT_PATH.relative_to(REPO_ROOT)}: {exc}'
-        ) from exc
+        _setup_failure(
+            f'typing_gate: cannot read {PYPROJECT_PATH.relative_to(REPO_ROOT)}: {exc}'
+        )
     except tomllib.TOMLDecodeError as exc:
-        raise SystemExit(
-            f'typing gate cannot parse {PYPROJECT_PATH.relative_to(REPO_ROOT)}: {exc}'
-        ) from exc
+        _setup_failure(
+            f'typing_gate: cannot parse {PYPROJECT_PATH.relative_to(REPO_ROOT)}: {exc}'
+        )
 
 
 def load_budget() -> dict[str, object]:
     if not BUDGET_PATH.exists():
-        raise SystemExit(
+        _setup_failure(
             f'typing gate cannot run: {BUDGET_PATH.relative_to(REPO_ROOT)} '
             f'is missing. Run with --update-budget to create it.'
         )
     try:
         return json.loads(BUDGET_PATH.read_text(encoding='utf-8'))
     except OSError as exc:
-        raise SystemExit(
-            f'typing gate cannot read {BUDGET_PATH.relative_to(REPO_ROOT)}: {exc}'
-        ) from exc
+        _setup_failure(
+            f'typing_gate: cannot read {BUDGET_PATH.relative_to(REPO_ROOT)}: {exc}'
+        )
     except json.JSONDecodeError as exc:
-        raise SystemExit(
-            f'typing gate cannot parse {BUDGET_PATH.relative_to(REPO_ROOT)}: {exc}'
-        ) from exc
+        _setup_failure(
+            f'typing_gate: cannot parse {BUDGET_PATH.relative_to(REPO_ROOT)}: {exc}'
+        )
 
 
 def main() -> int:
--- a/tools/typing_gate.py
+++ b/tools/typing_gate.py
@@ -863,8 +863,17 @@ def update_budget(pyright_json_path: str | None) -> None:
 # -------------------------------------------------------------------
 
 def load_pyproject() -> dict[str, object]:
-    with open(PYPROJECT_PATH, 'rb') as f:
-        return tomllib.load(f)
+    try:
+        with open(PYPROJECT_PATH, 'rb') as f:
+            return tomllib.load(f)
+    except OSError as exc:
+        raise SystemExit(
+            f'typing gate cannot read {PYPROJECT_PATH.relative_to(REPO_ROOT)}: {exc}'
+        ) from exc
+    except tomllib.TOMLDecodeError as exc:
+        raise SystemExit(
+            f'typing gate cannot parse {PYPROJECT_PATH.relative_to(REPO_ROOT)}: {exc}'
+        ) from exc
 
 
 def load_budget() -> dict[str, object]:
@@ -873,7 +882,16 @@ def load_budget() -> dict[str, object]:
             f'typing gate cannot run: {BUDGET_PATH.relative_to(REPO_ROOT)} '
             f'is missing. Run with --update-budget to create it.'
         )
-    return json.loads(BUDGET_PATH.read_text())
+    try:
+        return json.loads(BUDGET_PATH.read_text(encoding='utf-8'))
+    except OSError as exc:
+        raise SystemExit(
+            f'typing gate cannot read {BUDGET_PATH.relative_to(REPO_ROOT)}: {exc}'
+        ) from exc
+    except json.JSONDecodeError as exc:
+        raise SystemExit(
+            f'typing gate cannot parse {BUDGET_PATH.relative_to(REPO_ROOT)}: {exc}'
+        ) from exc
 
 
 def main() -> int:
@@ -937,14 +955,8 @@ def main() -> int:
         )
         return 1
 
-    try:
-        config = load_pyproject()
-        budget = load_budget()
-    except SystemExit:
-        raise
-    except Exception as e:
-        print(f'typing gate setup failed: {e}', file=sys.stderr)
-        return 2
+    config = load_pyproject()
+    budget = load_budget()
 
     failures: list[tuple[str, str]] = []
 
--- a/.github/rulesets/main.json
+++ b/.github/rulesets/main.json
@@ -74,6 +74,10 @@
           {
             "context": "pr_checks_ruleset",
             "integration_id": 15368
+          },
+          {
+            "context": "pr_checks_lint",
+            "integration_id": 15368
           }
         ]
       }
--- a/.github/workflows/pr_checks_lint.yml
+++ b/.github/workflows/pr_checks_lint.yml
@@ -1,0 +1,27 @@
+name: pr_checks_lint
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  pr_checks_lint:
+    name: pr_checks_lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - name: Set up lint venv
+        run: python -m venv .venv-lint
+      - name: Install Ruff
+        run: .venv-lint/bin/python -m pip install --upgrade pip 'ruff==0.15.11'
+      - name: Run Ruff lint gate
+        run: .venv-lint/bin/python -m ruff check tools tests/tools
--- a/.github/workflows/pr_checks_ruleset.yml
+++ b/.github/workflows/pr_checks_ruleset.yml
@@ -22,11 +22,14 @@ jobs:
         with:
           python-version: '3.11'
 
-      - name: Install test dependency
-        run: python -m pip install --upgrade pip pytest
+      - name: Set up contract venv
+        run: python -m venv .venv-ruleset
+
+      - name: Install contract dependencies
+        run: .venv-ruleset/bin/python -m pip install --upgrade pip pytest 'ruff==0.15.11'
 
       - name: Run CI contract tests
-        run: pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py -q
+        run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py -q
 
       - name: Compare live ruleset to checked-in snapshot
         env:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v1.3.3 on April 22, 2026
+- Add `.github/workflows/pr_checks_lint.yml` so `tools` and `tests/tools` gain a required fail-loud Ruff gate on `main`, pinned to Ruff `0.15.11`.
+- Extend `pr_checks_ruleset` with `tests/tools/test_lint_ci_contract.py` and a pinned Ruff install so the lint gate itself is mechanically protected by required CI.
+- Remove dead Ruff ignores `ANN101` and `ANN102`, add exact `BLE001` exemptions for `tools/slice_gate.py` and `tools/typing_gate.py`, and replace the remaining `RUF005` list concatenations in `tools/fail_loud_gate.py`.
+
 # v1.3.2 on April 22, 2026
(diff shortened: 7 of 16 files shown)


mikkokotila (commented):
One code-level finding remains inline on `pyproject.toml`: the new `BLE001` exemptions are file-wide, so `pr_checks_lint` will not catch fresh blind-except additions in `tools/slice_gate.py` or `tools/typing_gate.py` even though the slice text describes preserving only the two existing hits.

pyproject.toml
```diff
@@ -105,12 +105,12 @@ select = [
 ]
 ignore = [
   "E501",   # line length handled by formatter
-  "ANN101", # self typing not needed
-  "ANN102", # cls typing not needed
 ]
 
 [tool.ruff.lint.per-file-ignores]
 "quickstart_etl_tests/**/*.py" = ["S101", "ANN", "BLE001"]
+"tools/slice_gate.py" = ["BLE001"]
```
mikkokotila: These `BLE001` exemptions are file-wide, not occurrence-specific, so they do not preserve the "exact two existing hits" boundary the issue and PR body describe. I reproduced this against the exact PR head by adding a brand-new `except Exception: pass` block to `tools/typing_gate.py` in a temp checkout and then running the PR's pinned command `ruff check tools tests/tools`; it still returned `All checks passed!` because the whole file is exempted. Exact fix: either add a companion ratchet that pins the current `BLE001` count in these two files, or stop claiming these are exact exemptions. As written, a future PR can add new blind exceptions in either file without tripping `pr_checks_lint`.

mikkokotila:
## PR review against the issue #164 slice spec

All 8 required checks green on head `48fa22a`, including the new `pr_checks_lint` and the extended `pr_checks_ruleset`. The operator pre-rollout window appears to have completed cleanly — `pr_checks_ruleset`'s drift gate passing means the live ruleset already has the 8-context target including `pr_checks_lint`, applied before the PR could go green. The slice body was also updated to add the three `tests/fixtures/github/ruleset_live_*.json` files to Surfaces, closing what would have been a scope drift between the spec and what this PR actually touches.

Implementation matches the spec faithfully:

- [`pr_checks_lint.yml`](https://github.com/Vaquum/tdw-control-plane/blob/48fa22ad534cb99627bbd080f73352c503051958/.github/workflows/pr_checks_lint.yml) tracks the Signature Diff exactly: venv, pinned `ruff==0.15.11`, no `continue-on-error`, runs against `tools tests/tools`.
- [`pr_checks_ruleset.yml`](https://github.com/Vaquum/tdw-control-plane/blob/48fa22ad534cb99627bbd080f73352c503051958/.github/workflows/pr_checks_ruleset.yml) is extended via venv + same pinned Ruff + adds `tests/tools/test_lint_ci_contract.py` to the pytest invocation. Wires the new contract test into a required CI path.
- [`tests/tools/test_lint_ci_contract.py`](https://github.com/Vaquum/tdw-control-plane/blob/48fa22ad534cb99627bbd080f73352c503051958/tests/tools/test_lint_ci_contract.py) declares `RUFF_VERSION: Final[str] = '0.15.11'` once and uses it in three tests (workflow content checks + the consistency check). Effective single-source-of-truth pattern — the test module's constant is the shared anchor; both workflows must match it.
- [`tools/fail_loud_gate.py`](https://github.com/Vaquum/tdw-control-plane/blob/48fa22ad534cb99627bbd080f73352c503051958/tools/fail_loud_gate.py) has both `RUF005` fixes (`failures + ['msg']` → `[*failures, 'msg']`) exactly as specified.
- `pyproject.toml` removes the dead `ANN101`/`ANN102` ignores and adds `tools/slice_gate.py` + `tools/typing_gate.py` per-file `BLE001` exemptions, preserving runtime semantics.
- The bad fixture has the inline "do not 'fix'" header comment, lives outside the gate's scope (`tests/fixtures/lint/` not in `tools tests/tools`), so its intentional brokenness doesn't break the gate.

Two small notes worth recording — neither blocks merge.

### Note 1 — `test_pinned_ruff_fails_on_known_bad_fixture` does not actually pin Ruff

The test name promises pinning, but the implementation invokes:

```python
subprocess.run([sys.executable, '-m', 'ruff', 'check', str(BAD_FIXTURE)], ...)
```

That uses whatever `ruff` is installed in `sys.executable`'s environment. In the `pr_checks_ruleset` CI runner that environment is the `.venv-ruleset` venv where `ruff==0.15.11` was just installed, so the test does run against pinned Ruff in CI. But run locally with a system Python that has Ruff `0.16.x` installed, the test exercises `0.16.x`. A future Ruff version that, say, demoted `E401` from default-on to default-off would change the bad fixture's diagnostic count and the test's outcome would diverge between CI (pinned, fails as expected) and local (unpinned, possibly different).

The test currently passes in CI because of the venv's pin, not because of anything inside the test. Two ways to make the contract honest:

- rename to `test_ruff_fails_on_known_bad_fixture` (drops the "pinned" claim), and rely on the venv setup in `pr_checks_ruleset.yml` for the version guarantee, or
- have the test itself create an isolated venv, install `ruff==RUFF_VERSION`, and invoke that venv's ruff binary. Heavier but matches the name.

I lean toward option 1 — the pinning lives in the workflow, the test asserts the gate's failure mode. Don't overload one test with two contracts.

### Note 2 — Ruff scope expansion would catch the bad fixture

The bad fixture at `tests/fixtures/lint/bad_imports.py` is intentionally outside `tools tests/tools` so the gate skips it. But Ruff's `exclude` config in `pyproject.toml` does not list `tests/fixtures/`. If a future slice broadens the gate scope (e.g., `ruff check .` once the 206-error baseline is paid down per the Risks bullet), the gate would suddenly fail on this fixture.

Cheapest insurance: add `tests/fixtures/**` to `[tool.ruff].exclude` in this PR (or in the future scope-expansion slice). Sibling-slice candidate; not blocking.

### Verdict

PR is execution-ready. Note 1 is naming hygiene; Note 2 is forward-compat for the eventual full-repo scope expansion the issue's Unlocks already names. Recommend merging once the `mergeable_state: blocked` clears (likely Copilot review or required review thread resolution, not a check failure).

mikkokotila:
## Second pass — Note 1 closed, plus a substantive scope upgrade

The new commit `4a93964 fix(ci): remove BLE001 exemptions from lint gate` does more than the title suggests. Three things landed:

### Note 1 from my prior review — closed

`test_pinned_ruff_fails_on_known_bad_fixture` now invokes `ruff --version` first and asserts it equals `RUFF_VERSION` before running the gate against the bad fixture:

```python
version = _run_ruff('--version')
assert version.returncode == 0, version.stderr
assert version.stdout.strip() == f'ruff {RUFF_VERSION}'
```

That makes the test's name honest end-to-end. CI passes because the `.venv-ruleset` setup installs the pinned version; a local contributor with a different Ruff fails clearly with `ruff X.Y.Z != ruff 0.15.11` rather than silently testing against the wrong version. Cleaner than the rename I'd suggested.

### Substantive upgrade — BLE001 exemptions removed via real catch-clause narrowing

The previous spec deferred catch-clause narrowing to a hypothetical `runtime_env` slice and recorded `tools/slice_gate.py` + `tools/typing_gate.py` as per-file `BLE001` ignores. The new commit instead:

- replaces `except Exception` in [`tools/slice_gate.py::main`](https://github.com/Vaquum/tdw-control-plane/blob/4a93964a84b08a4a2a7fd0c560e66405b73219bf/tools/slice_gate.py) with direct `read_lines()` calls; `read_lines` itself now does explicit `print` + `raise SystemExit(2)` so the exit code semantics line up with the gate's documented contract (2 = setup failure)
- replaces `except Exception` in [`tools/typing_gate.py::main`](https://github.com/Vaquum/tdw-control-plane/blob/4a93964a84b08a4a2a7fd0c560e66405b73219bf/tools/typing_gate.py) with `try/except OSError + TOMLDecodeError/JSONDecodeError` inside `load_pyproject` and `load_budget`
- removes both per-file ignores from `pyproject.toml`
- the slice spec body was updated to reflect the new scope: Surfaces lists both files in Modified, two new Signature Diff entries spell out the literal Before/After, and the Risks bullet now reads *"the two broad-exception sites are replaced with explicit fail-loud read/setup handling"* instead of the previous "preserves runtime semantics" framing

This is the right call — actually fixing BLE001 instead of carrying forward an exemption — and the scope expansion is honestly disclosed in the spec.

### New regression protection: `test_pyproject_ruff_ignore_contract` now uses `--isolated`

```python
for path in ('tools/slice_gate.py', 'tools/typing_gate.py'):
    result = _run_ruff('check', '--isolated', '--select', 'BLE001', path)
    assert result.returncode == 0, result.stdout + result.stderr
```

`--isolated` tells Ruff to ignore `pyproject.toml`, so the test verifies BLE001 is clean on the file's own merits. A future PR cannot silently hide a new BLE001 violation by adding it back as a per-file-ignore. That is the right shape for this kind of ratchet.

### Tiny note (not blocking, not for this PR)

The narrower catch in `typing_gate.py::main` no longer wraps the `load_pyproject()` / `load_budget()` calls in a broad `except Exception → return 2`. Inside those two functions you handle the common cases (OSError, TOMLDecodeError, JSONDecodeError) with `SystemExit`. Esoteric exceptions (MemoryError on a huge pyproject, etc.) now propagate uncaught with a stack trace and Python's default exit code 1 instead of the prior clean `typing gate setup failed: <msg>` + return 2. Acceptable trade-off — the cases that occur in practice are covered, and the code is much clearer — but worth recording if anyone audits exit-code consistency across gates later.

### CI status

All 8 required checks green on head `4a93964`. `mergeable_state` is now `clean`. PR is ready to merge.

(Note 2 from my prior review — adding `tests/fixtures/**` to `[tool.ruff].exclude` for forward-compat with full-repo scope expansion — remains a sibling-slice candidate, not in scope here.)

tests/tools/test_lint_ci_contract.py
```diff
@@ -0,0 +1,93 @@
...
+    })
+
+    assert pins == [RUFF_VERSION]
+
+
+def test_pyproject_ruff_ignore_contract() -> None:
+    data = tomllib.loads((REPO_ROOT / 'pyproject.toml').read_text(encoding='utf-8'))
+    lint = data['tool']['ruff']['lint']
+
+    assert sorted(lint['ignore']) == ['E501']
```
mikkokotila: [P1] The contract still does not pin the actual Ruff policy surface. This test only fixes `ignore == ["E501"]`, forbids two specific per-file ignores, and proves one `I001` fixture still fails. The same PR can still weaken the law by editing other Ruff config and keep both required contexts green. Repro against the exact head `4a93964a84b08a4a2a7fd0c560e66405b73219bf`: in a temp copy, remove `"BLE"` from `[tool.ruff.lint].select`, add a new `except Exception: pass` block to `tools/fail_loud_gate.py`, then run the PR-pinned commands; `ruff check tools tests/tools` still prints `All checks passed!`, and `pytest tests/tools/test_lint_ci_contract.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_cc_gate.py -q` still passes `15 passed`. Exact fix: assert the authoritative Ruff config here (`select`, `exclude`, and the full allowed `per-file-ignores` surface), or run `pr_checks_lint` under `--isolated` with an explicit rule set so the candidate PR cannot relax its own policy.

mikkokotila:
## Third pass — `3f38c0c fix(ci): pin lint policy contract` is a real ratchet upgrade

The new commit reshapes [`test_pyproject_ruff_ignore_contract`](https://github.com/Vaquum/tdw-control-plane/blob/3f38c0c71f3ea57fe2ba8a03b995486e2ad477d3/tests/tools/test_lint_ci_contract.py) → `test_pyproject_ruff_policy_contract` and replaces the targeted "ignore is ['E501'] + two specific paths absent from per-file-ignores" assertions with a single literal-dict comparison against `EXPECTED_RUFF_POLICY`. The dict pins:

- `exclude` (5 named directories, exact order)
- `select` (7 rule categories: E, F, I, UP, RUF, BLE, ANN)
- `ignore` (just `['E501']`)
- `per-file-ignores` (only the `quickstart_etl_tests/**/*.py` entry)

Plus the BLE001 isolated check is now broadened to `ruff check --isolated --select BLE001 tools tests/tools` — the whole gated scope, not just the two previously-exempted files.

This is the right shape: a single literal contract that catches additions, removals, reorderings, and per-file scoping changes in one place. A future PR cannot quietly weaken the lint policy by adding a category to `ignore`, dropping a category from `select`, or scoping in a new per-file exemption — the contract test fails with a clear diff against the literal expected dict.

### Two small notes (neither blocking, neither for this PR)

1. **The contract pins `[tool.ruff].exclude`, `[tool.ruff.lint].select`, `[tool.ruff.lint].ignore`, and `[tool.ruff.lint.per-file-ignores]`, but does not pin `[tool.ruff].target-version`, `[tool.ruff].line-length`, `[tool.ruff.format]`, or `[tool.ruff.lint.isort]`.** Those affect what Ruff considers acceptable but are outside the policy contract today. If a future PR changed `target-version` from `py311` to `py310` (changing which `UP` rules apply) or `line-length` from `100` to `80` (changing which `E501` lines are reported — though `E501` is ignored, so moot), the contract test would not catch it. The test name is `test_pyproject_ruff_POLICY_contract` — defensible to say format/isort are out of scope — but worth recording for completeness.

2. **The pin is on rule categories (`E`, `F`, `I`, `UP`, `RUF`, `BLE`, `ANN`), not specific rule codes.** When the Ruff version eventually bumps in a future slice (per the Risks bullet), new rules added under any of those categories become enforced automatically. That is the intended behavior for a category pin and is the right default — but it makes the eventual bump-Ruff slice non-trivial because it has to deal with whatever new rules surface. Already named in Risks; flagging only because the policy contract makes the implication concrete.

### CI status

All 8 required checks green on head `3f38c0c`. `mergeable_state: blocked` is from new commits invalidating prior approvals, not a check failure.

### Verdict

PR is ready to merge. Each iteration of this PR has tightened the contract — pinning version, adding negative-path proof, narrowing catch clauses, now pinning the entire Ruff policy. The slice's principle ("hard, loud failing gates that govern everything they have capability for handling") is genuinely embodied here.

tools/typing_gate.py
```diff
@@ -863,8 +863,17 @@ def update_budget(pyright_json_path: str | None) -> None:
 # -------------------------------------------------------------------
 
 def load_pyproject() -> dict[str, object]:
-    with open(PYPROJECT_PATH, 'rb') as f:
-        return tomllib.load(f)
+    try:
+        with open(PYPROJECT_PATH, 'rb') as f:
+            return tomllib.load(f)
+    except OSError as exc:
+        raise SystemExit(
```
mikkokotila: [P2] These new helper-level `SystemExit(<message>)` paths regress `typing_gate` setup failures from exit code `2` to exit code `1`. On `origin/main`, read/parse failures from `load_pyproject()` and `load_budget()` were still caught by the outer `except Exception` path in `main()` and returned `2` as a hard setup error. On this head, they bypass that path entirely. Repro against current head `3f38c0c71f3ea57fe2ba8a03b995486e2ad477d3`: in a temp copy, corrupt `pyproject.toml` with `not = [valid`, then run `python3 tools/typing_gate.py --pyright-json /tmp/missing-pyright.json --bootstrap`; it exits `1` and prints only `typing gate cannot parse pyproject.toml: ...` on stderr. Exact fix: preserve the setup-error contract by printing the message to stderr and exiting `2` here (or restore the outer exception-to-`2` mapping) rather than raising `SystemExit(str)`.

mikkokotila:
## Fourth pass — `f8ff8ba fix(ci): restore typing gate setup exits` closes the second-pass note

The new commit addresses the exit-code-consistency concern I raised in pass two. Three things landed:

### `typing_gate.py` setup failures now exit cleanly with code 2 + stderr message

A new `_setup_failure(message)` helper prints to stderr and `raise SystemExit(2)`. Both `load_pyproject` and `load_budget` are rewritten to call it instead of `raise SystemExit(message)` (which Python exits with code 1, the wrong code for "gate setup failure" per the gates' shared contract).

```python
def _setup_failure(message: str) -> None:
    print(message, file=sys.stderr)
    raise SystemExit(2)
```

Message prefixes also normalized from `'typing gate cannot ...'` to `'typing_gate: cannot ...'` for consistency with `slice_gate:` and `cc_gate:`.

### New test pins the contract

[`test_typing_gate_setup_failures_exit_2`](https://github.com/Vaquum/tdw-control-plane/blob/f8ff8bae1ab68992424a7bc8c2d3ca7a736d42e3/tests/tools/test_ci_contract.py) creates a temp repo copy with deliberately invalid TOML in `pyproject.toml`, runs `typing_gate.py`, and asserts `returncode == 2`, `stdout == ''`, and the expected message on stderr. Exact pin on the failure mode.

### CHANGELOG updated honestly

The v1.3.3 entry is rewritten to drop *"add exact BLE001 exemptions"* and replace with *"replace broad-exception handling ... with explicit fail-loud setup/read handling"* — matching the actual scope.

### Tiny notes (neither blocking)

1. **The test pins one error path, not all four.** `_setup_failure` is called from four sites:
   - `load_pyproject`: OSError, TOMLDecodeError
   - `load_budget`: OSError, JSONDecodeError
   
   The new test only exercises `load_pyproject`'s TOMLDecodeError. The other three paths use the same helper so they should behave identically, but the test doesn't prove it. A parameterized test or three sibling tests would close the loop. Low value vs. test maintenance cost; defensible to leave as-is.

2. **The original "esoteric exceptions" concern is still live.** `typing_gate.py::main` no longer wraps `load_pyproject()` / `load_budget()` calls in a broad `except Exception → return 2`. Common cases (OSError, decode errors) are handled inside via the new helper. Esoteric exceptions (MemoryError on a huge pyproject, etc.) propagate uncaught with stack trace and Python's default exit 1. This is the deliberate trade-off — clean specific paths over broad-except defensiveness — and is consistent across the gate ecosystem now. Worth recording for whoever audits exit-code consistency later.

### CI status

Six required checks green on head `f8ff8ba`. `pr_checks_typing` and `PR Checks CodeQL (python)` still in flight (this commit pushed at 10:31 UTC). The pr_checks_typing run exercises the gate the commit just changed — happy-path behavior is unchanged so it should pass. Will confirm green once they complete.

### Verdict

PR is genuinely ready to merge once the two pending checks land green. Each iteration has tightened the contract: pin version → add negative-path proof → narrow catch clauses → pin entire Ruff policy → pin setup-failure exit code. The slice has gone through five honest revisions and the result is much sharper than the first PR I reviewed.

PR #166, opened 2026-04-22
feat: add Origo daily spot trades raw table, ingestion ledger, and pr_checks_tests gate

### LINKED ISSUE #157: feat: add Origo daily spot trades raw table, ingestion ledger, and pr_checks_tests gate

mikkokotila:
Closes #157

## Summary
- move the runtime image to Python 3.11 and replace the Origo path with a daily-source-native Binance spot trades template
- rename the Origo raw table surface to `binance_daily_spot_trades`, add the `binance_daily_spot_trades_ingestion` ledger, and make setup/ingest fail loud with explicit Dagster dependencies
- add the ClickHouse-backed `tests/origo_source_native` suite, fixture-backed Binance daily archives, `pr_checks_tests.yml`, and the checked-in ruleset snapshot requiring `pr_checks_tests`

## Local Proof
- `/tmp/tdw-origo-venv311/bin/python -m pytest tests/origo_source_native -q`
- `/tmp/tdw-origo-venv311/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py -q`
- `/tmp/tdw-origo-venv311/bin/python -m compileall tdw_control_plane tests tools`


mikkokotila:
## PR review against the issue #157 slice spec

This is the largest of the recent CI slices — 22 files, +1046/−383, real ClickHouse container in CI, full Origo path rewrite. Implementation is substantive and the tests are robust. Three things to flag and a fourth observation about the rollout state.

### Status — `pr_checks_ruleset` failure is expected, but the PR body doesn't say so

Live ruleset on `main` has 8 contexts; the snapshot in this PR adds `pr_checks_tests` for 9. Per the rollout sequence inherited from #160, the operator pre-applies the snapshot to ruleset 5406599 *before* the PR can go green. That hasn't happened yet, so [`pr_checks_ruleset`](https://github.com/Vaquum/tdw-control-plane/blob/$(git rev-parse origin/wa/origo-daily-spot-trades-main)/tools/ruleset_gate.py) reports drift correctly.

Worth one line in the PR body: *"`pr_checks_ruleset` will report drift until the operator pre-applies the 9-context snapshot per the #160/#164 rollout sequence; do not interpret this as a failing gate."* Otherwise a reviewer skimming Checks sees the red mark and may push back on a fundamentally healthy PR.

### Finding 1 — `pr_checks_tests` doesn't pin its dependencies

[`.github/workflows/pr_checks_tests.yml`](https://github.com/Vaquum/tdw-control-plane/blob/wa/origo-daily-spot-trades-main/.github/workflows/pr_checks_tests.yml) installs:

```yaml
- name: Install package and test dependencies
  run: python -m pip install --upgrade pip '.[dev]'
```

`'.[dev]'` resolves to whatever pip picks for `dagster`, `dagster-cloud`, `clickhouse-driver`, `clickhouse_connect`, `clickhouse-cityhash`, `dagster-webserver`, and `pytest` at install time. None are pinned in `pyproject.toml`.

This is inconsistent with the venv + pin pattern that #160 (pinned `pytest`) and #164 (pinned `ruff==0.15.11`) established. A future Dagster release that changes `materialize` semantics, or a clickhouse-driver release that changes `Client` behavior, would surface as a CI flake — loud, not silent, so the gate's principle is upheld. But the surface area for unintentional drift is much larger than the other two gates.

Two paths forward, neither blocking this PR:

- accept the trade-off explicitly in #157's Risks (the integration tests need a *real* runtime, and pinning `dagster`/`clickhouse-driver` etc. across the package is a different scope)
- pin via a constraints file or `pyproject.toml` lower bounds + upper bounds, plus a `test_pr_checks_tests_pins_dependencies` consistency check matching the #164 pattern

I lean toward the former — over-pinning runtime deps slows down legitimate dependency upgrades — but the lack of pinning is worth naming somewhere.

### Finding 2 — `dump_ruleset_contexts()` in conftest looks like dead code

[`tests/origo_source_native/conftest.py` lines 285-293](https://github.com/Vaquum/tdw-control-plane/blob/wa/origo-daily-spot-trades-main/tests/origo_source_native/conftest.py#L285-L293):

```python
def dump_ruleset_contexts() -> str:
    path = REPO_ROOT / '.github' / 'rulesets' / 'main.json'
    payload = json.loads(path.read_text(encoding='utf-8'))
    checks = next(
        rule['parameters']['required_status_checks']
        for rule in payload['rules']
        if rule['type'] == 'required_status_checks'
    )
    return '\n'.join(entry['context'] for entry in checks)
```

This function reads the ruleset snapshot and dumps required-status contexts. It is defined as a module-level function (not a fixture, not a test), is in the **origo_source_native** conftest (which is supposed to handle ClickHouse + Binance fixtures, not ruleset state), and grepping the test files shows no caller. Looks like leftover scaffolding from an earlier iteration — should either be removed or moved to `tests/tools/test_lint_ci_contract.py` if some test was supposed to use it.

### Finding 3 — heavyweight integration setup is fine, but `--maxfail=1` may hide cascading issues

The conftest spins up a real ClickHouse container per session via `docker run`, hosts the Binance daily fixture archives over a `ThreadingHTTPServer`, drops/recreates the `origo` database between tests. That is the right shape for proving the schema and the daily-ingest contract — integration tests for an integration system. Container is `--rm` + `--tmpfs` so cleanup is reliable.

Workflow runs `pytest tests/origo_source_native -q --maxfail=1`. The `--maxfail=1` stops at the first failure. For a fast-feedback dev loop that's right; for a CI gate proving the *full* contract on every PR, a single failure can mask several others, and the next PR that fixes the first failure may surface the second one as a fresh red. Worth considering `--maxfail=3` or dropping the cap so CI shows the full failure surface in one run.

Not a blocker; ergonomics.

### What works

- The implementation matches #157's MVC: idempotent database/table creation, declared Dagster asset deps (`create_binance_daily_spot_trades_table_origo` depends on `create_origo_database`; `insert_daily_binance_spot_trades_to_origo` depends on the table), exact source-row matching, ingestion ledger with checksums + counts + run metadata, fail-loud on missing `CLICKHOUSE_PASSWORD`, reingest-preserves-state.
- Dockerfile bumped to `python:3.11.12` (matches `test_dockerfile_uses_python_311`).
- Real Binance daily archive fixtures (zip + CHECKSUM) under `tests/fixtures/binance/spot/daily/trades/BTCUSDT/`. The CHECKSUM files prove integrity end-to-end through the ingest path.
- Workflow has `workflow_dispatch:` for manual re-runs.
- All other 9 required checks green: `pr_checks_typing`, `pr_checks_cc`, `pr_checks_slice`, `pr_checks_fail_loud`, `pr_checks_version`, `pr_checks_lint`, `pr_checks_tests` (the new one — runs against itself), `CodeQL`, `PR Checks CodeQL (python)`.

### Verdict

PR is ready to merge once the operator pre-applies the 9-context snapshot to live ruleset 5406599. None of the three findings block merge; finding 2 (dead `dump_ruleset_contexts`) is the cheapest one to address in this PR (single-line removal), the others are records for future slices.

.github/rulesets/main.json
```diff
@@ -78,6 +78,10 @@
           {
             "context": "pr_checks_lint",
             "integration_id": 15368
+          },
+          {
+            "context": "pr_checks_tests",
```
mikkokotila, line 83: [P1] The PR body says the live `Protect-Main` ruleset was updated before opening this rollout PR, but that is not true right now. The current live ruleset `5406599` still requires only eight contexts and does **not** include `pr_checks_tests`; that is exactly why `pr_checks_ruleset` is red on this PR today. Until the live ruleset is updated, this checked-in snapshot overclaims enforcement and the new test gate does not actually block merge at the repo level. Exact fix: apply the live ruleset update first, then rerun this PR so the snapshot and live state match.

.github/workflows/pr_checks_tests.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Install package and test dependencies
+        run: python -m pip install --upgrade pip '.[dev]'
+
+      - name: Run Origo source-native runtime tests
```
mikkokotila, line 28: [P1] `pr_checks_tests` is added as a required context, but nothing in required CI pins what this workflow must actually run. `pr_checks_ruleset` still executes only `test_cc_gate.py`, `test_ruleset_gate.py`, `test_ci_contract.py`, and `test_lint_ci_contract.py`; none of them inspect `.github/workflows/pr_checks_tests.yml`. I reproduced the bypass against the exact PR head in a temp copy by changing this step to `run: true` and then re-running the PR's own ruleset-contract suite: `pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py -q` still returned `16 passed`. So the same PR can weaken `pr_checks_tests` and keep `pr_checks_ruleset` green. Exact fix: add a dedicated CI-contract test that asserts this workflow's pinned command and forbids lax options like `continue-on-error`, then make `pr_checks_ruleset` run it.

tdw_control_plane/assets/create_origo_database.py
```diff
@@ -1,110 +1,69 @@
...
-        result = client.execute(f"SHOW DATABASES LIKE '{CLICKHOUSE_DATABASE}'")
-        if not result:
-            # Create the database with optimal settings
-            client.execute(f"""
-                CREATE DATABASE IF NOT EXISTS {CLICKHOUSE_DATABASE}
-                ENGINE = Atomic
-            """)
+def _get_clickhouse_settings() -> ClickHouseSettings:
+    return ClickHouseSettings(
+        host=_require_env('CLICKHOUSE_HOST'),
```
mikkokotila: [P0] This hard-requires env vars that the repo's own runtime surfaces do not set. Both `docker-compose.yml` and `docker-compose.deploy.yml` provide `CLICKHOUSE_PASSWORD`, but neither provides `CLICKHOUSE_HOST` or `CLICKHOUSE_PORT`; the pre-PR code defaulted those to `clickhouse` and `9000`. On the exact PR head, I reproduced the break in a temp checkout with a scratch venv: unset `CLICKHOUSE_HOST`, `CLICKHOUSE_PORT`, and `CLICKHOUSE_USER`, leave only `CLICKHOUSE_PASSWORD=x`, import `_get_clickhouse_settings()`, and it immediately raises `RuntimeError: CLICKHOUSE_HOST environment variable must be set.` That means the new `create_origo_database` / Origo setup path fails in the repo's own compose environment before it ever talks to ClickHouse. Exact fix: either preserve the old defaults for host/port/user here, or update every authoritative runtime surface to export those vars and prove that contract in tests.

mikkokotila:
## Second pass — substantive fixes plus a real production bug catch

Two new commits (`acb9989 fix: address Origo review findings` + `55ea9ee fix: satisfy Ruff lint contract`) address most of what I flagged and fix one real bug worth calling out separately.

### What landed

- **Production bug fix:** [`tdw_control_plane/definitions.py L304`](https://github.com/Vaquum/tdw-control-plane/blob/55ea9eef58d64e9dbbc2cafcbb5dfd4ae59bdef7/tdw_control_plane/definitions.py#L304) — the daily schedule referenced `insert_daily_binance_trades_job` (a stale name from the pre-rewrite Origo path); now correctly references `insert_daily_binance_spot_trades_to_origo_job`. Without this fix the scheduled daily ingest would have crashed at runtime trying to resolve a job that no longer exists. Catching it pre-merge is exactly what `pr_checks_tests` exists for.

- **Compose-default ergonomics restored:** `_get_clickhouse_settings()` now uses `os.environ.get(...)` defaults for HOST (`'clickhouse'`), PORT (`9000`), USER (`'default'`), and only PASSWORD remains required-with-`_require_env`. Plus a new `_get_clickhouse_port()` helper that converts to int with explicit `RuntimeError` on bad input. New test `test_origo_clickhouse_settings_preserve_compose_defaults` pins the contract.

- **Ledger payload tightened:** `_process_day` now records `source_file` (the zip filename, e.g., `BTCUSDT-trades-2024-01-01.zip`) and `csv_file` (the inner CSV name) as separate ledger fields instead of conflating them. More audit information.

- **Finding 1 (no dependency pinning) — partially addressed.** [`tests/tools/test_tests_ci_contract.py`](https://github.com/Vaquum/tdw-control-plane/blob/55ea9eef58d64e9dbbc2cafcbb5dfd4ae59bdef7/tests/tools/test_tests_ci_contract.py) pins the workflow's *content*: Python 3.11, the `'.[dev]'` install command, the `pytest tests/origo_source_native -q --maxfail=1` invocation, and the absence of `continue-on-error`. The new test is wired into `pr_checks_ruleset.yml`'s pytest invocation, mirroring the #164 contract pattern. So the workflow itself can no longer drift silently.

  This is the same pin-the-workflow-content approach #164 used for Ruff. The actual *dependency versions* (dagster, clickhouse-driver, dagster-cloud, etc.) remain unpinned in `pyproject.toml`, which I think is the right call for runtime deps but is now worth one Risks bullet in #157 to make the deliberate trade-off explicit.

### Still open from my prior review

- **Finding 2 (dead code `dump_ruleset_contexts` in `conftest.py`)** — not addressed. Function is still at [`tests/origo_source_native/conftest.py L285-293`](https://github.com/Vaquum/tdw-control-plane/blob/55ea9eef58d64e9dbbc2cafcbb5dfd4ae59bdef7/tests/origo_source_native/conftest.py#L285-L293) with no callers, in the wrong conftest's scope. One-line removal.

- **Finding 3 (`--maxfail=1` may hide cascading failures)** — now *pinned by* the new contract test. Changing it would require updating both the workflow and `EXPECTED_TEST_COMMAND` in `test_tests_ci_contract.py`. Not blocking; just noting that the test makes my suggestion harder to follow up on, ironically.

- **PR body still doesn't mention the expected `pr_checks_ruleset` failure.** A reader skimming Checks sees red on the most-relevant gate without context. Worth one line: *"`pr_checks_ruleset` will report drift until the operator pre-applies the 9-context snapshot per the #160/#164 rollout sequence; this is expected, not a failing gate."*

### CI status

8 of 9 unique required checks green on head `55ea9ee`. `pr_checks_ruleset` still failing — same drift-against-live-API state, same expected behavior per the rollout sequence. The new `test_pr_checks_tests_pins_python_and_runtime_suite_command` and `test_pr_checks_tests_workflow_exists` both pass inside the pytest step (verified by the green `pr_checks_lint`, etc., which install the same test set), so the workflow's content is now mechanically protected against future regressions even though the drift gate itself stays red until rollout.

### Verdict

PR is ready to merge once the operator applies the 9-context snapshot. The schedule-job-rename catch alone justifies the depth of review this slice has been getting.

mikkokotila:
## Third pass — `558cbc9 fix: harden Origo test harness imports` is a real cleanup, plus operator pre-rollout has happened

The new commit substantively improves the test harness architecture and addresses my outstanding finding 2. Operator has also pre-applied the 9-context snapshot — verified via `gh api repos/Vaquum/tdw-control-plane/rulesets/5406599`, live now has `["PR Checks CodeQL (python)", "pr_checks_typing", "pr_checks_cc", "pr_checks_slice", "pr_checks_fail_loud", "pr_checks_version", "pr_checks_ruleset", "pr_checks_lint", "pr_checks_tests"]`. `pr_checks_ruleset` is now `success` on the PR head.

### What landed in `558cbc9`

**Dead code removed.** `dump_ruleset_contexts()` is gone from `tests/origo_source_native/conftest.py` — closes finding 2.

**Import-time side effects eliminated.** The four `os.environ.setdefault(...)` calls at conftest module-load time are gone. They were masking a real coupling: the asset modules' `_get_clickhouse_settings()` would fail at import time without those defaults, so the conftest had to set them globally before importing the assets. The new architecture moves asset imports inside an `origo_assets` fixture that runs *after* `origo_test_env` has monkeypatched the env vars per-test. Cleaner separation of concerns and no global env pollution from importing the test module.

**Helper deduplication.** `load_expected_trade_rows` and `load_expected_ledger_payload` were defined in both `conftest.py` and `helpers.py` — now only in `helpers.py`. Single source of truth.

**`_wait_for_clickhouse` broad except narrowed** to four specific types (`OSError`, `RuntimeError`, `ValueError`, `EOFError`). Same lint-discipline pattern as the typing_gate / slice_gate narrowing in #165. Worth noting that `clickhouse-driver` can raise `clickhouse_driver.errors.ServerException` and other library-specific subclasses during connection attempts — none of those inherit from the four caught types. If clickhouse-driver ever changes which exception class fires during `SELECT 1` on a still-warming server, the test would crash with an uncaught exception instead of retrying. Acceptable for now (the four types cover today's common cases) but worth recording if anyone audits library exception coverage later.

**Timestamp parsing made strict.** Was *"if length 13 → ms, else → us"* — silently misinterpreted any non-13-digit value. Now explicit: `13 → ms, 16 → us, anything else → raise ValueError(...)`. Defensive improvement; catches a class of bug that would have produced subtly-wrong dates.

### My findings recap

- **Finding 1 (no dependency pinning)** — addressed in earlier commit via `test_pr_checks_tests_pins_python_and_runtime_suite_command`. Workflow content is now mechanically pinned. Actual dep versions still float; deliberate trade-off, worth one Risks bullet in #157.
- **Finding 2 (dead `dump_ruleset_contexts`)** — closed in this commit.
- **Finding 3 (`--maxfail=1`)** — pinned by the new contract test, would require coordinated update; not changing.
- **PR body still doesn't mention the (now-historical) `pr_checks_ruleset` failure context** — moot now that the operator has pre-applied and the gate is green. Worth one line in `#157`'s Closeout recording the bootstrap commands used, mirroring #160/#164.

### CI status

6 of 9 unique required checks green on head `558cbc9`. Three in flight at time of this comment: `pr_checks_typing`, `pr_checks_tests`, `PR Checks CodeQL (python)`. None expected to fail (the test harness changes don't touch what those gates check, except `pr_checks_tests` which exercises the very test suite that was reorganized — the green `pr_checks_ruleset` already confirms the new contract test passes).

### Verdict

PR is essentially merge-ready. Once the three in-flight checks complete green, this can land. The slice has gone through a notable evolution: initial implementation → review-driven hardening → real bug catch (the schedule job rename) → architectural cleanup (lazy imports, no global env pollution) → operator rollout. Five honest iterations, much sharper than the first commit.

mikkokotila:
## Final pass — all green, mergeable

All 9 unique required checks now `success` on head `558cbc9`:

- `PR Checks CodeQL (python)` ✓
- `pr_checks_typing` ✓
- `pr_checks_cc` ✓
- `pr_checks_slice` ✓
- `pr_checks_fail_loud` ✓
- `pr_checks_version` ✓
- `pr_checks_ruleset` ✓
- `pr_checks_lint` ✓
- `pr_checks_tests` ✓

`mergeable_state: clean`. Ready to merge.

When you record the closeout in #157, the bootstrap commands the operator used to apply the 9-context snapshot to ruleset 5406599 should be captured (mirroring #160's and #164's closeout records) so the rollout sequence is auditable end-to-end.

Issue #168, opened 2026-04-22
feat(origo): complete binance spot new data source template


mikkokotila:
## Slice ID

S06

## Slice Title

feat(origo): complete binance spot new data source template

## Routing

> **Significance.** Routing is the minimum information needed to place this
> slice in the dependency graph. Parent and dependencies must be existing
> issue numbers; unlocks may be forward-looking. Prose description of the
> relationship is not accepted — only issue references. This section is not
> a planning document; it is a graph edge list.

Parent issue: none
Depends on:
- #157
Unlocks:

## Minimum Viable Capability

> **Significance.** A slice delivers exactly one capability. The capability
> must be expressible as a set of assertions that become mechanically true
> at merge. This is the answer to *what changes in the system that an
> external observer can verify by running a command?* Every line below
> must be evaluable as a shell command, SQL query, Python expression, or
> HTTP request that returns a definite result. One assertion per line.
>
> Accepted:
> - `clickhouse-client --query "SELECT countIf(stream_sequence IS NULL) FROM canonical_event_log"` returns `0`.
> - `python -c "import inspect, origo.events.writer as m; print(inspect.signature(m.CanonicalEventWriter.__init__))"` returns `(self, *, clickhouse_client)`.
> - `curl -sf https://origo.vaquum.fi/v1/health/schema | jq -r '.canonical_event_log_v1.columns | length'` returns `17`.
>
> Rejected:
> - *Freeze the current-truth read contract.*
> - *Strict CQRS separation.*
> - *Terminal proof obligations are complete.*

Assertions (one per line, each executable as-is):
- `python3 -c "import importlib; m = importlib.import_module('tdw_control_plane.assets.create_binance_spot_klines_table_origo'); print(m.KLINES_TABLE_NAME)"` returns `binance_spot_klines`
- `python3 -c "import importlib; m = importlib.import_module('tdw_control_plane.assets.create_aligned_1m_exchange_table_origo'); print(m.ALIGNED_TABLE_NAME)"` returns `aligned_1m_exchange`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_daily_pipeline_schedule_targets_binance_spot_data_source_job -q` exits `0`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_binance_spot_klines_schema_matches_exchange_contract -q` exits `0`
[...]

PR #173, opened 2026-04-23
fix(origo): make Binance spot projection schema and refresh SQL match TDW

### LINKED ISSUE #172: fix(origo): make Binance spot projection schema and refresh SQL match TDW
## Slice ID

S08

## Slice Title

fix(origo): make Binance spot projection schema and refresh SQL match TDW

## Routing

Parent issue: none
Depends on:
- #168
Unlocks:
- (none)

## Minimum Viable Capability

>
> Accepted:
>
> Rejected:
> - *Strict CQRS separation.*

- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_binance_spot_klines_schema_matches_tdw_contract_fixture -q` exits `0`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_binance_spot_klines_rows_match_tdw_contract_fixture_for_fixture_day -q` exits `0`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_aligned_1m_exchange_schema_matches_tdw_contract_fixture_plus_dataset_source -q` exits `0`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_aligned_1m_exchange_rows_match_tdw_contract_fixture_plus_dataset_source -q` exits `0`

## Current State

>
> Accepted:
>
> Rejected:

- `python3 -c "from pathlib import Path; t = Path('tdw_control_plane/assets/create_binance_spot_klines_table_origo.py').read_text(); print('close_time UInt64' in t, 'mean Float64' in t, 'maker_ratio Float64' in t)"` returns `True False False`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_binance_spot_klines_exact_rows_match_expected -q` exits `0`
- `python3 -c "from pathlib import Path; t = Path('tdw_control_plane/assets/create_aligned_1m_exchange_table_origo.py').read_text(); print('dataset_source LowCardinality(String)' in t, 'close_time UInt64' in t, 'mean Float64' in t)"` returns `True True False`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_aligned_1m_exchange_rows_from_binance_spot_match_expected -q` exits `0`
- `test -f tests/fixtures/tdw/binance_spot_1m_contract.json; echo $?` returns `1`

## Target State

> predicate.
>
> Accepted:

- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_binance_spot_klines_schema_matches_tdw_contract_fixture -q` exits `0`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_binance_spot_klines_rows_match_tdw_contract_fixture_for_fixture_day -q` exits `0`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_aligned_1m_exchange_schema_matches_tdw_contract_fixture_plus_dataset_source -q` exits `0`
- `python3 -m pytest tests/origo_source_native/test_origo_binance_spot_data_source_template.py::test_aligned_1m_exchange_rows_match_tdw_contract_fixture_plus_dataset_source -q` exits `0`
- `python3 -c "from pathlib import Path; t = Path('tdw_control_plane/assets/create_binance_spot_klines_table_origo.py').read_text(); print('close_time UInt64' in t, 'mean Float64' in t, 'maker_ratio Float64' in t)"` returns `False True True`
- `python3 -c "from pathlib import Path; t = Path('tdw_control_plane/assets/create_aligned_1m_exchange_table_origo.py').read_text(); print('dataset_source LowCardinality(String)' in t, 'close_time UInt64' in t, 'mean Float64' in t)"` returns `True False True`
- `test -f tests/fixtures/tdw/binance_spot_1m_contract.json; echo $?` returns `0`

## Surfaces

- `tdw_control_plane/assets/create_binance_spot_klines_table_origo.py`
- `tdw_control_plane/assets/refresh_binance_spot_klines_origo.py`
- `tdw_control_plane/assets/create_aligned_1m_exchange_table_origo.py`
- `tdw_control_plane/assets/refresh_aligned_1m_exchange_from_binance_spot_origo.py`
- `tests/origo_source_native/test_origo_binance_spot_data_source_template.py`
- `pyproject.toml`
- `CHANGELOG.md`

Added:
- `tests/fixtures/tdw/binance_spot_1m_contract.json`

Removed:
- (none)

## Signature Diff

### tdw_control_plane/assets/create_binance_spot_klines_table_origo.py::CREATE TABLE
Before:
```sql
CREATE TABLE IF NOT EXISTS origo.binance_spot_klines (
    open_time UInt64,
    open Float64,
    high Float64,
    low Float64,
    close Float64,
    volume Float64,
    close_time UInt64,
    quote_asset_volume Float64,
    number_of_trades UInt64,
    taker_buy_base_asset_volume Float64,
    taker_buy_quote_asset_volume Float64,
    ignore Float64
)
ENGINE = MergeTree()
PARTITION BY toYYYYMM(fromUnixTimestamp64Milli(open_time))
ORDER BY open_time
```
After:
```sql
CREATE TABLE IF NOT EXISTS origo.binance_spot_klines (
    datetime DateTime,
    open Float64,
    high Float64,
    low Float64,
    close Float64,
    mean Float64,
    std Float64,
    median Float64,
    iqr Float64,
    volume Float64,
    maker_ratio Float64,
    no_of_trades UInt64,
    open_liquidity Float64,
    high_liquidity Float64,
    low_liquidity Float64,
    close_liquidity Float64,
    liquidity_sum Float64,
    maker_volume Float64,
    maker_liquidity Float64
)
ENGINE = MergeTree()
PARTITION BY toYYYYMM(datetime)
ORDER BY datetime
```

### tdw_control_plane/assets/refresh_binance_spot_klines_origo.py::projection SQL
Before:
```sql
INSERT INTO {database}.binance_spot_klines
SELECT
    toUnixTimestamp(toStartOfMinute(datetime)) * 1000 AS open_time,
    argMin(price, trade_id) AS open,
    max(price) AS high,
    min(price) AS low,
    argMax(price, trade_id) AS close,
    sum(quantity) AS volume,
    ((toUnixTimestamp(toStartOfMinute(datetime)) + 60) * 1000) - 1 AS close_time,
    sum(quote_quantity) AS quote_asset_volume,
    count() AS number_of_trades,
    sumIf(quantity, is_buyer_maker = 0) AS taker_buy_base_asset_volume,
    sumIf(quote_quantity, is_buyer_maker = 0) AS taker_buy_quote_asset_volume,
    toFloat64(0) AS ignore
FROM {database}.binance_daily_spot_trades
WHERE toDate(datetime) = toDate('{partition_date}')
GROUP BY toStartOfMinute(datetime)
ORDER BY open_time
```
After:
```sql
INSERT INTO {database}.binance_spot_klines
SELECT
    kline_datetime AS datetime,
    argMin(price, trade_id) AS open,
    max(price) AS high,
    min(price) AS low,
    argMax(price, trade_id) AS close,
    avg(price) AS mean,
    stddevPopStable(price) AS std,
    quantileExact(0.5)(price) AS median,
    quantileExact(0.75)(price) - quantileExact(0.25)(price) AS iqr,
    sumKahan(quantity) AS volume,
    avg(is_buyer_maker) AS maker_ratio,
    count() AS no_of_trades,
    argMin(price * quantity, trade_id) AS open_liquidity,
    max(price * quantity) AS high_liquidity,
    min(price * quantity) AS low_liquidity,
    argMax(price * quantity, trade_id) AS close_liquidity,
    sum(price * quantity) AS liquidity_sum,
    sumKahan(is_buyer_maker * quantity) AS maker_volume,
    sum(is_buyer_maker * price * quantity) AS maker_liquidity
FROM (
    SELECT
        *,
        toDateTime(60 * intDiv(toUnixTimestamp(datetime), 60)) AS kline_datetime
    FROM {database}.binance_daily_spot_trades
    WHERE toDate(datetime) = toDate('{partition_date}')
)
GROUP BY kline_datetime
ORDER BY kline_datetime
```

### tdw_control_plane/assets/create_aligned_1m_exchange_table_origo.py::CREATE TABLE
Before:
```sql
CREATE TABLE IF NOT EXISTS origo.aligned_1m_exchange (
    dataset_source LowCardinality(String),
    open_time UInt64,
    open Float64,
    high Float64,
    low Float64,
    close Float64,
    volume Float64,
    close_time UInt64,
    quote_asset_volume Float64,
    number_of_trades UInt64,
    taker_buy_base_asset_volume Float64,
    taker_buy_quote_asset_volume Float64,
    ignore Float64
)
ENGINE = MergeTree()
PARTITION BY toYYYYMM(fromUnixTimestamp64Milli(open_time))
ORDER BY (dataset_source, open_time)
```
After:
```sql
CREATE TABLE IF NOT EXISTS origo.aligned_1m_exchange (
    dataset_source LowCardinality(String),
    datetime DateTime,
    open Float64,
    high Float64,
    low Float64,
    close Float64,
    mean Float64,
    std Float64,
    median Float64,
    iqr Float64,
    volume Float64,
    maker_ratio Float64,
    no_of_trades UInt64,
    open_liquidity Float64,
    high_liquidity Float64,
    low_liquidity Float64,
    close_liquidity Float64,
    liquidity_sum Float64,
    maker_volume Float64,
    maker_liquidity Float64
)
ENGINE = MergeTree()
PARTITION BY toYYYYMM(datetime)
ORDER BY (dataset_source, datetime)
```

### tdw_control_plane/assets/refresh_aligned_1m_exchange_from_binance_spot_origo.py::projection SQL
Before:
```sql
INSERT INTO {database}.aligned_1m_exchange
SELECT
    'binance_spot' AS dataset_source,
    open_time,
    open,
    high,
    low,
    close,
    volume,
    close_time,
    quote_asset_volume,
    number_of_trades,
    taker_buy_base_asset_volume,
    taker_buy_quote_asset_volume,
    ignore
FROM {database}.binance_spot_klines
WHERE toDate(fromUnixTimestamp64Milli(open_time)) = toDate('{partition_date}')
ORDER BY open_time
```
After:
```sql
INSERT INTO {database}.aligned_1m_exchange
SELECT
    'binance_spot' AS dataset_source,
    datetime,
    open,
    high,
    low,
    close,
    mean,
    std,
    median,
    iqr,
    volume,
    maker_ratio,
    no_of_trades,
    open_liquidity,
    high_liquidity,
    low_liquidity,
    close_liquidity,
    liquidity_sum,
    maker_volume,
    maker_liquidity
FROM {database}.binance_spot_klines
WHERE toDate(datetime) = toDate('{partition_date}')
ORDER BY datetime
```

Before:
```toml
version = "1.5.0"
```
After:
```toml
version = "1.5.1"
```

### tests/fixtures/tdw/binance_spot_1m_contract.json::root payload
Before:
```json
(does not exist)
```
After:
```json
{
  "columns": ["datetime", "open", "high", "low", "close", "mean", "std", "median", "iqr", "volume", "maker_ratio", "no_of_trades", "open_liquidity", "high_liquidity", "low_liquidity", "close_liquidity", "liquidity_sum", "maker_volume", "maker_liquidity"],
  "fixture_day": "<YYYY-MM-DD>",
  "binance_spot_klines_rows": [{"datetime": "...", "open": 0.0, "high": 0.0, "low": 0.0, "close": 0.0, "mean": 0.0, "std": 0.0, "median": 0.0, "iqr": 0.0, "volume": 0.0, "maker_ratio": 0.0, "no_of_trades": 0, "open_liquidity": 0.0, "high_liquidity": 0.0, "low_liquidity": 0.0, "close_liquidity": 0.0, "liquidity_sum": 0.0, "maker_volume": 0.0, "maker_liquidity": 0.0}],
  "aligned_1m_exchange_rows": [{"dataset_source": "binance_spot", "datetime": "...", "open": 0.0, "high": 0.0, "low": 0.0, "close": 0.0, "mean": 0.0, "std": 0.0, "median": 0.0, "iqr": 0.0, "volume": 0.0, "maker_ratio": 0.0, "no_of_trades": 0, "open_liquidity": 0.0, "high_liquidity": 0.0, "low_liquidity": 0.0, "close_liquidity": 0.0, "liquidity_sum": 0.0, "maker_volume": 0.0, "maker_liquidity": 0.0}]
}
```

Before:
```markdown
# v1.5.0 on April 23, 2026
```
After:
```markdown
# v1.5.1 on April 23, 2026
```

## Tests That Prove the MVC

|---|---|---|
| `binance_spot_klines` schema matches the TDW contract fixture | `tests/origo_source_native/test_origo_binance_spot_data_source_template.py` | `test_binance_spot_klines_schema_matches_tdw_contract_fixture` |
| `binance_spot_klines` rows match the TDW contract fixture for the fixture day | `tests/origo_source_native/test_origo_binance_spot_data_source_template.py` | `test_binance_spot_klines_rows_match_tdw_contract_fixture_for_fixture_day` |
| `aligned_1m_exchange` schema matches the TDW contract fixture plus `dataset_source` | `tests/origo_source_native/test_origo_binance_spot_data_source_template.py` | `test_aligned_1m_exchange_schema_matches_tdw_contract_fixture_plus_dataset_source` |
| `aligned_1m_exchange` rows match the TDW contract fixture plus `dataset_source` | `tests/origo_source_native/test_origo_binance_spot_data_source_template.py` | `test_aligned_1m_exchange_rows_match_tdw_contract_fixture_plus_dataset_source` |

## Regression Gate

>
> Accepted:
>   on non-zero exit.
>
> Rejected:

Workflow file:
Job name:
- `pr_checks_tests`

## Out of Scope

- `tdw_control_plane/query/**`
- `.github/workflows/**`
- `.github/rulesets/**`
- `tests/tools/**`

## Risks and Open Questions

Risks:
- This slice is the smallest honest corrective slice. It changes the table schemas and the refresh SQL together so the live refresh path stays valid.
- This slice must not change jobs, schedules, raw insert behavior, or projection refresh orchestration.
- The live downstream TDW consumer path remains out of scope. This slice makes the Origo schema and refresh output match the checked-in TDW contract fixture for Binance spot.
- The checked-in TDW contract fixture is a slice-time snapshot derived from `tdw_control_plane/query/get_binance_spot_klines.py`. Future TDW contract changes require a sibling corrective slice that updates both the fixture and the Origo projection surfaces.
- The fixture Signature Diff pins the JSON root shape and key ordering. The actual `fixture_day` and row values live in `tests/fixtures/tdw/binance_spot_1m_contract.json` and are derived from the existing Binance fixture archive.
- Existing non-fresh Origo projection tables with the old schema must be dropped before the corrected CREATE TABLE assets can take effect. Exact operator commands: `clickhouse-client --query "DROP TABLE IF EXISTS origo.binance_spot_klines SYNC"` and `clickhouse-client --query "DROP TABLE IF EXISTS origo.aligned_1m_exchange SYNC"`.
- If the actual TDW contract differs from the fixture chosen in this issue, the issue must be corrected before implementation.
- The existing `test_*_exact_rows_match_expected` row tests are removed and replaced by the new TDW-contract-fixture row tests in this slice.

Open questions:
- None.

## Closeout

> bot verifies:

Merge SHA:
Merged PR number:
- `pr_checks_tests`:
- `pr_checks_version`:

## Author Checks

mikkokotila:
Closes #172

## What changed
- replace the Origo `binance_spot_klines` and `aligned_1m_exchange` table schemas with the TDW 1-minute kline contract
- replace both projection refresh SQL paths so they materialize the TDW analytics columns from Binance spot raw trades
- add a checked-in TDW contract fixture and replace the old exchange-native row tests with fixture-backed schema and row-parity tests
- bump the package version to `1.5.1` and record the corrective slice in `CHANGELOG.md`

## Proof
- `/tmp/tdw-168-venv311/bin/python -m pytest tests/origo_source_native -q --maxfail=1`
- `/tmp/tdw-168-venv311/bin/python -m compileall tdw_control_plane tests tools`
- `/tmp/tdw-168-venv311/bin/python tools/version_gate.py --pr-title 'fix(origo): make Binance spot projection schema and refresh SQL match TDW' --base-pyproject /tmp/tdw-172-base-pyproject.toml --head-pyproject pyproject.toml --base-changelog /tmp/tdw-172-base-CHANGELOG.md --head-changelog CHANGELOG.md`

### DIFF
--- a/tests/fixtures/tdw/binance_spot_1m_contract.json
+++ b/tests/fixtures/tdw/binance_spot_1m_contract.json
@@ -1,0 +1,71 @@
+{
+  "columns": [
+    "datetime",
+    "open",
+    "high",
+    "low",
+    "close",
+    "mean",
+    "std",
+    "median",
+    "iqr",
+    "volume",
+    "maker_ratio",
+    "no_of_trades",
+    "open_liquidity",
+    "high_liquidity",
+    "low_liquidity",
+    "close_liquidity",
+    "liquidity_sum",
+    "maker_volume",
+    "maker_liquidity"
+  ],
+  "fixture_day": "2024-01-01",
+  "binance_spot_klines_rows": [
+    {
+      "datetime": "2024-01-01 00:00:00",
+      "open": 42000.1,
+      "high": 42010.0,
+      "low": 42000.1,
+      "close": 42005.5,
+      "mean": 42005.200000000004,
+      "std": 4.047221268968605,
+      "median": 42005.5,
+      "iqr": 9.900000000001455,
+      "volume": 0.0045000000000000005,
+      "maker_ratio": 0.6666666666666666,
+      "no_of_trades": 3,
+      "open_liquidity": 42.000099999999996,
+      "high_liquidity": 84.02,
+      "low_liquidity": 42.000099999999996,
+      "close_liquidity": 63.008250000000004,
+      "liquidity_sum": 189.02835,
+      "maker_volume": 0.0025,
+      "maker_liquidity": 105.00835000000001
+    }
+  ],
+  "aligned_1m_exchange_rows": [
+    {
+      "dataset_source": "binance_spot",
+      "datetime": "2024-01-01 00:00:00",
+      "open": 42000.1,
+      "high": 42010.0,
+      "low": 42000.1,
+      "close": 42005.5,
+      "mean": 42005.200000000004,
+      "std": 4.047221268968605,
+      "median": 42005.5,
+      "iqr": 9.900000000001455,
+      "volume": 0.0045000000000000005,
+      "maker_ratio": 0.6666666666666666,
+      "no_of_trades": 3,
+      "open_liquidity": 42.000099999999996,
+      "high_liquidity": 84.02,
+      "low_liquidity": 42.000099999999996,
+      "close_liquidity": 63.008250000000004,
+      "liquidity_sum": 189.02835,
+      "maker_volume": 0.0025,
+      "maker_liquidity": 105.00835000000001
+    }
+  ]
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v1.5.1 on April 23, 2026
+- Correct the Origo Binance spot projection contract so `binance_spot_klines` and `aligned_1m_exchange` match the TDW 1-minute kline schema instead of the exchange-native 12-column shape.
+- Replace the single-source and aligned refresh SQL so both tables materialize the TDW analytics columns (`mean`, `std`, `median`, `iqr`, maker/liquidity fields) from raw spot trades.
+- Add a checked-in TDW contract fixture and replace the old exchange-native row tests with fixture-backed schema and row-parity tests for both projection tables.
+
 # v1.5.0 on April 23, 2026
 - Complete the first Binance spot Origo data-source template on top of `binance_daily_spot_trades` by adding the single-source `binance_spot_klines` projection and the shared `aligned_1m_exchange` projection layer.
 - Replace the old raw-only daily Origo schedule target with `refresh_binance_spot_data_source_job`, which materializes the raw daily insert plus both projection layers for the same partition.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "quickstart_etl"
-version = "1.5.0"
+version = "1.5.1"
 description = "Trades Warehouse Control Plane"
 readme = "README.md"
 requires-python = ">=3.11"
--- a/tdw_control_plane/assets/create_aligned_1m_exchange_table_origo.py
+++ b/tdw_control_plane/assets/create_aligned_1m_exchange_table_origo.py
@@ -17,22 +17,29 @@ def _create_aligned_table(client: ClickhouseClient, settings: ClickHouseSettings
         f"""
         CREATE TABLE IF NOT EXISTS {settings.database}.{ALIGNED_TABLE_NAME} (
             dataset_source LowCardinality(String),
-            open_time UInt64,
+            datetime DateTime,
             open Float64,
             high Float64,
             low Float64,
             close Float64,
+            mean Float64,
+            std Float64,
+            median Float64,
+            iqr Float64,
             volume Float64,
-            close_time UInt64,
-            quote_asset_volume Float64,
-            number_of_trades UInt64,
-            taker_buy_base_asset_volume Float64,
-            taker_buy_quote_asset_volume Float64,
-            ignore Float64
+            maker_ratio Float64,
+            no_of_trades UInt64,
+            open_liquidity Float64,
+            high_liquidity Float64,
+            low_liquidity Float64,
+            close_liquidity Float64,
+            liquidity_sum Float64,
+            maker_volume Float64,
+            maker_liquidity Float64
         )
         ENGINE = MergeTree()
-        PARTITION BY toYYYYMM(fromUnixTimestamp64Milli(open_time))
-        ORDER BY (dataset_source, open_time)
+        PARTITION BY toYYYYMM(datetime)
+        ORDER BY (dataset_source, datetime)
         """
     )
 
--- a/tdw_control_plane/assets/create_binance_spot_klines_table_origo.py
+++ b/tdw_control_plane/assets/create_binance_spot_klines_table_origo.py
@@ -16,22 +16,29 @@ def _create_klines_table(client: ClickhouseClient, settings: ClickHouseSettings)
     client.execute(
         f"""
         CREATE TABLE IF NOT EXISTS {settings.database}.{KLINES_TABLE_NAME} (
-            open_time UInt64,
+            datetime DateTime,
             open Float64,
             high Float64,
             low Float64,
             close Float64,
+            mean Float64,
+            std Float64,
+            median Float64,
+            iqr Float64,
             volume Float64,
-            close_time UInt64,
-            quote_asset_volume Float64,
-            number_of_trades UInt64,
-            taker_buy_base_asset_volume Float64,
-            taker_buy_quote_asset_volume Float64,
-            ignore Float64
+            maker_ratio Float64,
+            no_of_trades UInt64,
+            open_liquidity Float64,
+            high_liquidity Float64,
+            low_liquidity Float64,
+            close_liquidity Float64,
+            liquidity_sum Float64,
+            maker_volume Float64,
+            maker_liquidity Float64
         )
         ENGINE = MergeTree()
-        PARTITION BY toYYYYMM(fromUnixTimestamp64Milli(open_time))
-        ORDER BY open_time
+        PARTITION BY toYYYYMM(datetime)
+        ORDER BY datetime
         """
     )
 
--- a/tdw_control_plane/assets/refresh_aligned_1m_exchange_from_binance_spot_origo.py
+++ b/tdw_control_plane/assets/refresh_aligned_1m_exchange_from_binance_spot_origo.py
@@ -28,7 +28,7 @@ def _delete_partition_rows(
         f"""
         ALTER TABLE {database}.{ALIGNED_TABLE_NAME}
         DELETE WHERE dataset_source = '{BINANCE_SPOT_DATASET_SOURCE}'
-          AND toDate(fromUnixTimestamp64Milli(open_time)) = toDate('{partition_date}')
+          AND toDate(datetime) = toDate('{partition_date}')
         """,
         settings={'mutations_sync': 2},
     )
@@ -44,7 +44,7 @@ def _count_partition_rows(
         SELECT count()
         FROM {database}.{ALIGNED_TABLE_NAME}
         WHERE dataset_source = '{BINANCE_SPOT_DATASET_SOURCE}'
-          AND toDate(fromUnixTimestamp64Milli(open_time)) = toDate('{partition_date}')
+          AND toDate(datetime) = toDate('{partition_date}')
         """
     )
     return int(result[0][0])
@@ -60,21 +60,28 @@ def _insert_partition_rows(
         INSERT INTO {database}.{ALIGNED_TABLE_NAME}
         SELECT
             '{BINANCE_SPOT_DATASET_SOURCE}' AS dataset_source,
-            open_time,
+            datetime,
             open,
             high,
             low,
             close,
+            mean,
+            std,
+            median,
+            iqr,
             volume,
-            close_time,
-            quote_asset_volume,
-            number_of_trades,
-            taker_buy_base_asset_volume,
-            taker_buy_quote_asset_volume,
-            ignore
+            maker_ratio,
+            no_of_trades,
+            open_liquidity,
+            high_liquidity,
+            low_liquidity,
+            close_liquidity,
+            liquidity_sum,
+            maker_volume,
+            maker_liquidity
         FROM {database}.{KLINES_TABLE_NAME}
-        WHERE toDate(fromUnixTimestamp64Milli(open_time)) = toDate('{partition_date}')
-        ORDER BY open_time
+        WHERE toDate(datetime) = toDate('{partition_date}')
+        ORDER BY datetime
         """
     )
 
--- a/tdw_control_plane/assets/refresh_binance_spot_klines_origo.py
+++ b/tdw_control_plane/assets/refresh_binance_spot_klines_origo.py
@@ -29,7 +29,7 @@ def _delete_partition_rows(
     client.execute(
         f"""
         ALTER TABLE {database}.{KLINES_TABLE_NAME}
-        DELETE WHERE toDate(fromUnixTimestamp64Milli(open_time)) = toDate('{partition_date}')
+        DELETE WHERE toDate(datetime) = toDate('{partition_date}')
         """,
         settings={'mutations_sync': 2},
     )
@@ -44,7 +44,7 @@ def _count_partition_rows(
         f"""
         SELECT count()
         FROM {database}.{KLINES_TABLE_NAME}
-        WHERE toDate(fromUnixTimestamp64Milli(open_time)) = toDate('{partition_date}')
+        WHERE toDate(datetime) = toDate('{partition_date}')
         """
     )
     return int(result[0][0])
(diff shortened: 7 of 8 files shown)


zero-bang (approved):
LGTM. Thorough review confirms:

- Column order is consistent across all four schema/refresh files, the fixture JSON, and the test type lists.
- The refresh SQL in `refresh_binance_spot_klines_origo.py` mirrors the production `get_binance_spot_klines.py` query pattern (same `toDateTime(60 * intDiv(...))` truncation, same aggregate functions, same `GROUP BY datetime` alias resolution).
- `publish_binance_spot_klines_to_huggingface.py` queries `tdw.binance_trades_complete` via the query module — unaffected by this Origo schema change.
- No remaining references to old column names (`open_time`, `close_time`, `quote_asset_volume`, etc.) in any production code path.
- Tests verify both schema shape and row parity against the checked-in fixture, and the idempotency test now compares first-run vs second-run output directly.
- All CI gates green.

Minor carry-forward: `helpers.py` still defines `KLINE_SCHEMA_COLUMNS`, `ALIGNED_SCHEMA_COLUMNS`, `load_expected_binance_spot_kline_rows`, and `load_expected_aligned_1m_exchange_rows` — these are now dead code after this PR removes the imports. Worth a cleanup pass, but not blocking.

mikkokotila:
## PR review against the rescoped #172 slice

Implementation tracks the slice spec faithfully. All 14 unique required checks green, including `pr_checks_tests` which exercises the row-shape parity tests against the new schema. Title sync between issue and PR is correct (`fix(origo): make Binance spot projection schema and refresh SQL match TDW` byte-equal in both places). The 8 files in the PR map exactly to the slice's Modified+Added Surfaces.

### What's right

- **Idempotency mechanism is now visible** in [`refresh_binance_spot_klines_origo.py L29-35`](https://github.com/Vaquum/tdw-control-plane/blob/wa/origo-binance-spot-tdw-kline-contract-main/tdw_control_plane/assets/refresh_binance_spot_klines_origo.py#L29-L35) and the parallel aligned refresh: `ALTER TABLE ... DELETE WHERE toDate(datetime) = toDate('{partition_date}') WITH SETTINGS mutations_sync=2`, then INSERT. This is the right ClickHouse pattern for partition replacement (sync mutation barrier before INSERT) and it closes the abstract idempotency concern I flagged on #168 — same-partition reruns now have a visible mechanism, not just an asserted property.
- **Time column migration is consistent end-to-end.** Old `open_time UInt64` (ms epoch) → new `datetime DateTime`. The DELETE/count predicates updated together with the INSERT, no `fromUnixTimestamp64Milli(open_time)` left over. That's the kind of consistency that the prior "columns only" rescope would have broken.
- **The fixture's row values are mechanically derived.** The kline row at [`binance_spot_1m_contract.json L24-45`](https://github.com/Vaquum/tdw-control-plane/blob/wa/origo-binance-spot-tdw-kline-contract-main/tests/fixtures/tdw/binance_spot_1m_contract.json#L24-L45) shows `mean: 42005.200000000004`, which is exactly `(42000.1 + 42010.0 + 42005.5) / 3 = 42005.2` to floating-point precision — confirming the fixture was computed from the projection SQL, not hand-written. The `aligned_1m_exchange_rows` is byte-identical to `binance_spot_klines_rows` except for the `dataset_source: "binance_spot"` prefix, validating the pure-copy semantics of the aligned projection.
- **Test reorganization is honest.** The slice's Risks bullet committed to removing `test_*_exact_rows_match_expected` and replacing with fixture-backed variants. The PR's `tests/origo_source_native/test_origo_binance_spot_data_source_template.py` shows +122/-90, consistent with that scope of replacement.

### Two semantic concerns carried through from the slice review

These were raised on the slice spec and are now PR-visible because the SQL is in the diff:

**Concern 1 — `sumKahan` vs `sum` asymmetry.** [`refresh_binance_spot_klines_origo.py L68-79`](https://github.com/Vaquum/tdw-control-plane/blob/wa/origo-binance-spot-tdw-kline-contract-main/tdw_control_plane/assets/refresh_binance_spot_klines_origo.py#L68-L79):

```sql
sumKahan(quantity) AS volume
sum(price * quantity) AS liquidity_sum
sumKahan(is_buyer_maker * quantity) AS maker_volume
sum(is_buyer_maker * price * quantity) AS maker_liquidity
```

Base-quantity sums use `sumKahan` (numerically stable); price-multiplied (liquidity) sums use regular `sum`. This is the wrong direction for floating-point accuracy — `price * quantity` products amplify magnitude before summing, so the liquidity sums benefit *more* from Kahan, not less.

If this asymmetry exactly matches TDW's `get_binance_spot_klines.py` implementation, that's defensible (the slice's Risks bullet explicitly names that file as the contract source of truth). If it's an oversight in either TDW or this slice, both should be revisited. A one-line note in the PR description or a Risks bullet in #172 stating *"the sumKahan/sum asymmetry mirrors TDW's existing implementation; both files inherit the same trade-off"* would close this.

**Concern 2 — `mean = avg(price)` is simple mean, not VWAP.** Confirmed by the fixture math: `mean: 42005.200000000004 = (42000.1 + 42010.0 + 42005.5) / 3`. For trading data the more standard "mean price per minute" is volume-weighted (`sum(price * quantity) / sum(quantity)`), which would give a different value because the larger trade at 42010.0 has weight 0.0025 vs 0.001 for the others.

If TDW's `get_binance_spot_klines.py` defines `mean` as simple `avg(price)`, this is correct. If it's VWAP, both this implementation and the fixture are wrong (and the test passes only because they're consistently wrong together). The fixture and the projection SQL were authored together, so the test gives no independent verification of the TDW contract.

The slice's Risks bullet acknowledges this trade-off: *"The checked-in TDW contract fixture is a slice-time snapshot derived from `tdw_control_plane/query/get_binance_spot_klines.py`."* But "derived from" leaves open whether the derivation was mechanical (read TDW's SQL, port to fixture) or interpretive (wrote the projection, generated the fixture). Worth one sentence in the PR description recording which.

### Forward-looking observation, not for this PR

The strongest contract test would compare this PR's projection output to TDW's `get_binance_spot_klines.py` output for the same input data, row-by-row, rather than comparing both to a hand-curated fixture. The fixture-based test catches drift between projection and fixture; it does not catch drift between either and TDW. A follow-up slice could add `tests/origo_source_native/test_binance_spot_klines_matches_tdw_query.py` that runs both code paths over the same fixture day and asserts byte-equal output. That would close the meta-concern from my first slice review.

### Verdict

PR is ready to merge. The structural reshape (schema + refresh SQL together, with the idempotency mechanism visible in code) is exactly what the slice promised. The two semantic concerns are not blocking — they are bets on whether the TDW contract source actually matches the implementation here, and either the fixture is right (in which case both surfaces agree) or it is wrong (in which case downstream consumers would notice quickly). Worth a brief PR-description note recording how the fixture was derived.

zero-bang (approved):

mikkokotila:
## Second pass — `9d56da2` catches a real SQL ambiguity bug

The new commit fixes a genuine correctness issue in the projection SQL — the `datetime` column alias was ambiguous between the source table's per-trade `datetime` and the bucketed minute `datetime` in `GROUP BY datetime` / `ORDER BY datetime`.

```sql
-- Before
SELECT
    toDateTime(60 * intDiv(toUnixTimestamp(datetime), 60)) AS datetime,
    ...
FROM {database}.{RAW_TABLE_NAME}
WHERE toDate(datetime) = toDate('{partition_date}')
GROUP BY datetime          -- ambiguous: source datetime or bucket?
ORDER BY datetime          -- same ambiguity
```

```sql
-- After
SELECT
    kline_datetime AS datetime,
    ...
FROM (
    SELECT *,
           toDateTime(60 * intDiv(toUnixTimestamp(datetime), 60)) AS kline_datetime
    FROM {database}.{RAW_TABLE_NAME}
    WHERE toDate(datetime) = toDate('{partition_date}')
)
GROUP BY kline_datetime    -- unambiguously the bucket
ORDER BY kline_datetime
```

The bucket gets a distinct alias inside a subquery; the outer SELECT renames it back to `datetime` for the INSERT target column. Clean.

This was a quietly dangerous bug. The fixture in `tests/fixtures/tdw/binance_spot_1m_contract.json` has 3 trades all on `2024-01-01 00:00:00` — they bucket to the same minute regardless of which `datetime` ClickHouse resolves, so the test wouldn't expose the ambiguity. In production, with a partition's worth of multi-minute trades, the `GROUP BY datetime` could have grouped by per-trade datetimes (no aggregation) instead of by minute buckets, silently producing one row per trade rather than one row per minute. Not the kind of bug that fails loudly.

### Test-coverage observation (not blocking, not for this PR)

The current fixture only exercises a single-minute scenario. A multi-minute fixture day (say, 30+ trades distributed across several different minutes) would have caught this `GROUP BY` ambiguity at test time. Worth recording as a follow-up — `tests/fixtures/tdw/binance_spot_1m_contract.json` could grow a `multi_minute_fixture_day` section with rows spanning multiple buckets, plus a sibling test that asserts row count matches the bucket count rather than the trade count. Catches an entire class of bucket-aggregation bugs that the single-minute fixture is blind to.

### Status

All 10 unique required checks green on head `9d56da2`. `mergeable_state: clean`. PR is ready to merge.

The two semantic concerns from my first PR review (`sumKahan` vs `sum` asymmetry, and `mean = avg(price)` vs VWAP) remain unaddressed but were not blocking — they are bets on whether the projection matches TDW's `get_binance_spot_klines.py`, and either the fixture is right (both surfaces agree) or it is wrong (downstream consumers will notice). The PR description recording how the fixture was derived (mechanical port from TDW vs. interpretive co-authoring) would still help future maintainers, but it's not on the merge critical path.

tests/fixtures/tdw/binance_spot_1m_contract.json
```diff
@@ -0,0 +1,71 @@
...
+  ],
+  "fixture_day": "2024-01-01",
+  "binance_spot_klines_rows": [
+    {
+      "datetime": "2024-01-01 00:00:00",
+      "open": 42000.1,
+      "high": 42010.0,
+      "low": 42000.1,
+      "close": 42005.5,
+      "mean": 42005.200000000004,
```
mikkokotila, line 31: [P1] This fixture no longer matches the TDW helper the slice names as source of truth. `get_binance_spot_klines()` rounds `mean`, `std`, `volume`, `liquidity_sum`, and `maker_liquidity` after fetching the query result ([tdw_control_plane/query/get_binance_spot_klines.py:143-155]), but this fixture pins the pre-rounding aggregate values instead (`42005.200000000004` vs `42005.2`, `4.047221268968605` vs `4.047221`, `189.02835` vs `189.0`, etc.). That means the new tests prove parity to a different contract than the slice says is authoritative. Exact fix: either generate the fixture from the helper's post-processed output, or explicitly narrow the claimed contract to the raw SQL aggregate output and stop naming `get_binance_spot_klines.py` as the source of truth.

PR #179, opened 2026-04-23
ci: add post-merge privileged ruleset parity audit for bypass_actors on main

### LINKED ISSUE #163: ci: add post-merge privileged ruleset parity audit for bypass_actors on main
## Slice ID

S11

## Slice Title

ci: add post-merge privileged ruleset parity audit for bypass_actors on main

## Routing

Parent issue: none
Depends on:
- #160
Unlocks:
- (none)

## Minimum Viable Capability

>
> Accepted:
>
> Rejected:
> - *Strict CQRS separation.*

- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_privileged_ruleset_audit_accepts_matching_live_payload_with_bypass_actors -q` exits `0`
- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_privileged_ruleset_audit_fails_on_bypass_actor_drift -q` exits `0`
- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_privileged_ruleset_audit_writes_live_payload_snapshot_on_failure -q` exits `0`
- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_audit_main_ruleset_workflow_contract -q` exits `0`
- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_privileged_ruleset_audit_fails_loud_when_live_payload_omits_bypass_actors -q` exits `0`

## Current State

>
> Accepted:
>
> Rejected:

- `test -f .github/workflows/audit_main_ruleset.yml; echo $?` returns `1`
- `test -f tools/privileged_ruleset_audit.py; echo $?` returns `1`
- `test -f tests/tools/test_privileged_ruleset_audit.py; echo $?` returns `1`
- `python3 -m pytest tests/tools/test_ruleset_gate.py::test_missing_bypass_actors_is_allowed_for_repo_hosted_reads -q` exits `0`

## Target State

> predicate.
>
> Accepted:

- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_privileged_ruleset_audit_accepts_matching_live_payload_with_bypass_actors -q` exits `0`
- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_privileged_ruleset_audit_fails_on_bypass_actor_drift -q` exits `0`
- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_privileged_ruleset_audit_writes_live_payload_snapshot_on_failure -q` exits `0`
- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_audit_main_ruleset_workflow_contract -q` exits `0`
- `python3 -m pytest tests/tools/test_privileged_ruleset_audit.py::test_privileged_ruleset_audit_fails_loud_when_live_payload_omits_bypass_actors -q` exits `0`

## Surfaces

- `pyproject.toml`
- `CHANGELOG.md`

Added:
- `.github/workflows/audit_main_ruleset.yml`
- `tools/privileged_ruleset_audit.py`
- `tests/tools/test_privileged_ruleset_audit.py`

Removed:
- (none)

## Signature Diff

### .github/workflows/audit_main_ruleset.yml::workflow trigger and privileged job
Before:
```yaml
# does not exist on origin/main
```
After:
```yaml
name: audit_main_ruleset

on:
  push:
    branches: [main]
  workflow_dispatch:

permissions:
  contents: read

jobs:
  audit_main_ruleset:
    if: github.ref == 'refs/heads/main'
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-python@v5
        with:
          python-version: '3.11'
      - run: python -m venv .venv-ruleset-audit
      - run: .venv-ruleset-audit/bin/python -m pip install --upgrade pip
      - env:
          GH_TOKEN: ${{ secrets.RULESET_AUDIT_TOKEN }}
        run: >-
          .venv-ruleset-audit/bin/python tools/privileged_ruleset_audit.py
          --ruleset-file .github/rulesets/main.json
          --repo "${{ github.repository }}"
          --ruleset-id 5406599
          --output-dir .ruleset-audit
```

### tools/privileged_ruleset_audit.py::CLI contract
Before:
```python
# does not exist on origin/main
```
After:
```python
parser.add_argument('--ruleset-file', required=True)
parser.add_argument('--repo', required=True)
parser.add_argument('--ruleset-id', required=True, type=int)
parser.add_argument('--output-dir', required=True)

def normalize_privileged_live_ruleset(payload: dict[str, object]) -> dict[str, object]:
    ...
```

### .github/workflows/pr_checks_ruleset.yml::CI contract test invocation
Before:
```yaml
- name: Run CI contract tests
  run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py -q
```
After:
```yaml
- name: Run CI contract tests
  run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py tests/tools/test_privileged_ruleset_audit.py -q
```

Before:
```toml
version = "1.6.6"
```
After:
```toml
version = "1.6.7"
```

Before:
```markdown
# v1.6.6 on April 30, 2026
```
After:
```markdown
# v1.6.7 on May 1, 2026
```

## Tests That Prove the MVC

|---|---|---|
| matching live payload with visible `bypass_actors` passes the privileged audit | `tests/tools/test_privileged_ruleset_audit.py` | `test_privileged_ruleset_audit_accepts_matching_live_payload_with_bypass_actors` |
| `bypass_actors` drift fails the privileged audit | `tests/tools/test_privileged_ruleset_audit.py` | `test_privileged_ruleset_audit_fails_on_bypass_actor_drift` |
| privileged audit writes the exact compared live payload on failure | `tests/tools/test_privileged_ruleset_audit.py` | `test_privileged_ruleset_audit_writes_live_payload_snapshot_on_failure` |
| privileged audit workflow runs on `main` push or guarded manual dispatch, never on candidate PR code paths | `tests/tools/test_privileged_ruleset_audit.py` | `test_audit_main_ruleset_workflow_contract` |
| privileged audit fails loud with setup error when the live payload omits `bypass_actors` | `tests/tools/test_privileged_ruleset_audit.py` | `test_privileged_ruleset_audit_fails_loud_when_live_payload_omits_bypass_actors` |

## Regression Gate

>
> Accepted:
>   on non-zero exit.
>
> Rejected:

Workflow file:
Job name:
- `pr_checks_ruleset`
- `.venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py tests/tools/test_privileged_ruleset_audit.py -q`

## Out of Scope

- `tools/ruleset_gate.py`
- `tdw_control_plane/**`
- `tests/origo_source_native/**`

## Risks and Open Questions

Risks:
- This slice is explicitly a privileged post-merge audit on `main`. It does not post a required status to the PR head SHA, and it does not convert `pr_checks_ruleset` into full-field pre-merge enforcement.
- The existing repo-hosted observable-subset downgrade in `tools/ruleset_gate.py` remains in place after this slice. This slice complements that path; it does not remove it.
- The privileged token must never be available to `pull_request` candidate code. The audit workflow must stay on `push: [main]` and `workflow_dispatch` with a job-level `github.ref == 'refs/heads/main'` guard only.
- The privileged audit path must verify that the live payload includes `bypass_actors` and fail loud with a setup error if `RULESET_AUDIT_TOKEN` is missing or underscoped. It must never silently fall back to the observable-subset comparison used by the repo-hosted PR path.
- The privileged audit must share normalization and equality semantics with `tools/ruleset_gate.py` rather than reimplementing a drifting second comparison law.
- If `audit_main_ruleset` fails on `main`, operator response is part of the slice contract: either restore the live ruleset to the checked-in snapshot or open a sibling slice that updates the snapshot before the next ruleset-affecting merge.
- If GitHub changes the ruleset API shape again, the privileged audit must fail loud rather than silently strip new fields.
- If future policy requires pre-merge privileged enforcement on the latest PR head SHA, that is a sibling slice with a different credential and status-publication design.
- Because the slice title is `ci(...)`, the governed version trail is a patch bump: `1.6.6 -> 1.6.7` after the current `main` advancement.

Open questions:
- None.

## Closeout

> bot verifies:

Merge SHA:
Merged PR number:
- `PR Checks CodeQL (python)`:
- `pr_checks_typing`:
- `pr_checks_cc`:
- `pr_checks_slice`:
- `pr_checks_fail_loud`:
- `pr_checks_version`:
- `pr_checks_ruleset`:
- `pr_checks_lint`:
- `pr_checks_tests`:
- `audit_main_ruleset`:

## Author Checks

mikkokotila:
Closes #163

## Summary
- Add a main-only privileged ruleset audit workflow using `RULESET_AUDIT_TOKEN`, with a job-level `github.ref == 'refs/heads/main'` guard for manual dispatch.
- Add the audit tool and contract tests, including fail-loud behavior when `bypass_actors` is not visible.
- Extend `pr_checks_ruleset` so the new workflow/tool contract is protected before rollout.
- Rebase-by-merge onto current `main` and bump version/changelog to `1.6.7`.

## Local checks
- `UV_PROJECT_ENVIRONMENT=/tmp/tdw-control-plane-uv-env /Users/mikkokotila/.langflow/uv/uv run --frozen --extra dev --with ruff==0.15.11 python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py tests/tools/test_privileged_ruleset_audit.py -q`
- `UV_PROJECT_ENVIRONMENT=/tmp/tdw-control-plane-uv-env /Users/mikkokotila/.langflow/uv/uv run --frozen --extra dev python -m pytest tests/origo_source_native -q --maxfail=1`
- `UV_PROJECT_ENVIRONMENT=/tmp/tdw-control-plane-uv-env /Users/mikkokotila/.langflow/uv/uv run --with ruff==0.15.11 ruff check tools tests/tools`
- `UV_PROJECT_ENVIRONMENT=/tmp/tdw-control-plane-uv-env /Users/mikkokotila/.langflow/uv/uv run --frozen --extra dev python tools/fail_loud_gate.py --base-budget .github/fail_loud_budget.json`
- `UV_PROJECT_ENVIRONMENT=/tmp/tdw-control-plane-uv-env /Users/mikkokotila/.langflow/uv/uv run --frozen --extra dev --with pyright==1.1.408 pyright --outputjson > /tmp/tdw-control-plane-pyright.json; ... tools/typing_gate.py --pyright-json /tmp/tdw-control-plane-pyright.json --base-budget .github/typing_budget.json` (`pyright` exits on existing config warning; typing gate passes)
- `tools/version_gate.py` against `origin/main` (`1.6.6 -> 1.6.7`)

### DIFF
--- a/.github/workflows/audit_main_ruleset.yml
+++ b/.github/workflows/audit_main_ruleset.yml
@@ -26,7 +26,7 @@ jobs:
         run: python -m venv .venv-ruleset-audit
 
       - name: Install audit dependencies
-        run: .venv-ruleset-audit/bin/python -m pip install --upgrade pip pytest
+        run: .venv-ruleset-audit/bin/python -m pip install --upgrade pip
 
       - name: Audit live main ruleset with privileged token
         env:
--- a/.github/workflows/audit_main_ruleset.yml
+++ b/.github/workflows/audit_main_ruleset.yml
@@ -1,0 +1,46 @@
+name: audit_main_ruleset
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  audit_main_ruleset:
+    name: audit_main_ruleset
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Set up audit venv
+        run: python -m venv .venv-ruleset-audit
+
+      - name: Install audit dependencies
+        run: .venv-ruleset-audit/bin/python -m pip install --upgrade pip pytest
+
+      - name: Audit live main ruleset with privileged token
+        env:
+          GH_TOKEN: ${{ secrets.RULESET_AUDIT_TOKEN }}
+        run: >-
+          .venv-ruleset-audit/bin/python tools/privileged_ruleset_audit.py
+          --ruleset-file .github/rulesets/main.json
+          --repo "${{ github.repository }}"
+          --ruleset-id 5406599
+          --output-dir .ruleset-audit
+
+      - name: Upload compared live payload on failure
+        if: ${{ failure() && hashFiles('.ruleset-audit/*') != '' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: audit-main-ruleset-live-payload
+          path: .ruleset-audit
--- a/tests/tools/test_privileged_ruleset_audit.py
+++ b/tests/tools/test_privileged_ruleset_audit.py
@@ -24,6 +24,15 @@ def _load_audit_module() -> types.ModuleType:
     return module
 
 
+def _load_ruleset_gate_module() -> types.ModuleType:
+    spec = importlib.util.spec_from_file_location('ruleset_gate_independent', REPO_ROOT / 'tools/ruleset_gate.py')
+    assert spec is not None
+    assert spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
 def _run_audit_fixture(live_fixture: str, tmp_path: Path) -> tuple[int, str, str]:
     module = _load_audit_module()
     stdout = io.StringIO()
@@ -33,7 +42,7 @@ def _run_audit_fixture(live_fixture: str, tmp_path: Path) -> tuple[int, str, str
             code = module.run_audit(
                 ruleset_file=str(SNAPSHOT),
                 repo=None,
-                ruleset_id='5406599',
+                ruleset_id=5406599,
                 output_dir=str(tmp_path),
                 live_json=str(FIXTURES / live_fixture),
             )
@@ -98,17 +107,18 @@ def test_privileged_ruleset_audit_fails_loud_when_live_payload_omits_bypass_acto
 
 def test_privileged_ruleset_audit_shares_ruleset_gate_semantics() -> None:
     module = _load_audit_module()
+    ruleset_gate = _load_ruleset_gate_module()
 
     assert (
         module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
-        == module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
+        == ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
     )
     assert (
         module.shared_ruleset_gate.OPTIONAL_LIVE_TOP_LEVEL_FIELDS
-        == frozenset({'bypass_actors'})
+        == ruleset_gate.OPTIONAL_LIVE_TOP_LEVEL_FIELDS
     )
-    assert 'bypass_actors' in module.shared_ruleset_gate.SNAPSHOT_TOP_LEVEL_FIELDS
-    assert 'bypass_actors' not in module.shared_ruleset_gate.IGNORED_LIVE_FIELDS
+    assert module.shared_ruleset_gate.SNAPSHOT_TOP_LEVEL_FIELDS == ruleset_gate.SNAPSHOT_TOP_LEVEL_FIELDS
+    assert module.shared_ruleset_gate.IGNORED_LIVE_FIELDS == ruleset_gate.IGNORED_LIVE_FIELDS
 
 
 def test_audit_main_ruleset_workflow_contract() -> None:
--- a/tests/tools/test_privileged_ruleset_audit.py
+++ b/tests/tools/test_privileged_ruleset_audit.py
@@ -1,0 +1,137 @@
+from __future__ import annotations
+
+import contextlib
+import importlib.util
+import io
+import json
+import types
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+AUDIT_TOOL = REPO_ROOT / 'tools/privileged_ruleset_audit.py'
+RULESET_WORKFLOW = REPO_ROOT / '.github/workflows/pr_checks_ruleset.yml'
+AUDIT_WORKFLOW = REPO_ROOT / '.github/workflows/audit_main_ruleset.yml'
+SNAPSHOT = REPO_ROOT / '.github/rulesets/main.json'
+FIXTURES = REPO_ROOT / 'tests/fixtures/github'
+
+
+def _load_audit_module() -> types.ModuleType:
+    spec = importlib.util.spec_from_file_location('privileged_ruleset_audit', AUDIT_TOOL)
+    assert spec is not None
+    assert spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def _run_audit_fixture(live_fixture: str, tmp_path: Path) -> tuple[int, str, str]:
+    module = _load_audit_module()
+    stdout = io.StringIO()
+    stderr = io.StringIO()
+    with contextlib.redirect_stdout(stdout), contextlib.redirect_stderr(stderr):
+        try:
+            code = module.run_audit(
+                ruleset_file=str(SNAPSHOT),
+                repo=None,
+                ruleset_id='5406599',
+                output_dir=str(tmp_path),
+                live_json=str(FIXTURES / live_fixture),
+            )
+        except SystemExit as exc:
+            code = int(exc.code)
+    return code, stdout.getvalue(), stderr.getvalue()
+
+
+def test_privileged_ruleset_audit_accepts_matching_live_payload_with_bypass_actors(
+    tmp_path: Path,
+) -> None:
+    code, stdout, stderr = _run_audit_fixture('ruleset_live_target.json', tmp_path)
+
+    assert code == 0
+    assert 'PRIVILEGED RULESET AUDIT -- PASS' in stdout
+    assert stderr == ''
+    assert not (tmp_path / 'live_ruleset.json').exists()
+
+
+def test_privileged_ruleset_audit_fails_on_bypass_actor_drift(tmp_path: Path) -> None:
+    code, _, stderr = _run_audit_fixture('ruleset_live_with_bypass_actor.json', tmp_path)
+
+    assert code == 1
+    assert 'privileged_ruleset_audit: ruleset drift detected' in stderr
+
+
+def test_privileged_ruleset_audit_writes_live_payload_snapshot_on_failure(
+    tmp_path: Path,
+) -> None:
+    code, _, stderr = _run_audit_fixture('ruleset_live_with_bypass_actor.json', tmp_path)
+    snapshot = tmp_path / 'live_ruleset.json'
+
+    assert code == 1
+    assert snapshot.exists()
+    assert json.loads(snapshot.read_text(encoding='utf-8')) == json.loads(
+        (FIXTURES / 'ruleset_live_with_bypass_actor.json').read_text(encoding='utf-8')
+    )
+    assert str(snapshot) in stderr
+
+
+def test_privileged_ruleset_audit_fails_loud_when_live_payload_omits_bypass_actors(
+    tmp_path: Path,
+) -> None:
+    code, _, stderr = _run_audit_fixture(
+        'ruleset_live_target_without_bypass_actors.json',
+        tmp_path,
+    )
+    snapshot = tmp_path / 'live_ruleset.json'
+
+    assert code == 2
+    assert (
+        "privileged live ruleset missing required observable field(s): ['bypass_actors']"
+        in stderr
+    )
+    assert snapshot.exists()
+    assert json.loads(snapshot.read_text(encoding='utf-8')) == json.loads(
+        (FIXTURES / 'ruleset_live_target_without_bypass_actors.json').read_text(
+            encoding='utf-8'
+        )
+    )
+
+
+def test_privileged_ruleset_audit_shares_ruleset_gate_semantics() -> None:
+    module = _load_audit_module()
+
+    assert (
+        module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
+        == module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
+    )
+    assert (
+        module.shared_ruleset_gate.OPTIONAL_LIVE_TOP_LEVEL_FIELDS
+        == frozenset({'bypass_actors'})
+    )
+    assert 'bypass_actors' in module.shared_ruleset_gate.SNAPSHOT_TOP_LEVEL_FIELDS
+    assert 'bypass_actors' not in module.shared_ruleset_gate.IGNORED_LIVE_FIELDS
+
+
+def test_audit_main_ruleset_workflow_contract() -> None:
+    workflow = AUDIT_WORKFLOW.read_text(encoding='utf-8')
+
+    assert 'name: audit_main_ruleset' in workflow
+    assert 'push:' in workflow
+    assert 'branches: [main]' in workflow
+    assert 'workflow_dispatch:' in workflow
+    assert 'pull_request:' not in workflow
+    assert 'RULESET_AUDIT_TOKEN' in workflow
+    assert 'tools/privileged_ruleset_audit.py' in workflow
+    assert '--ruleset-file .github/rulesets/main.json' in workflow
+    assert '--repo "${{ github.repository }}"' in workflow
+    assert '--ruleset-id 5406599' in workflow
+    assert '--output-dir .ruleset-audit' in workflow
+    assert 'actions/upload-artifact@v4' in workflow
+    assert "failure()" in workflow
+    assert '.ruleset-audit' in workflow
+
+
+def test_pr_checks_ruleset_runs_privileged_audit_contract() -> None:
+    workflow = RULESET_WORKFLOW.read_text(encoding='utf-8')
+
+    assert 'tests/tools/test_privileged_ruleset_audit.py' in workflow
+    assert 'continue-on-error' not in workflow
--- a/.github/workflows/pr_checks_ruleset.yml
+++ b/.github/workflows/pr_checks_ruleset.yml
@@ -29,7 +29,7 @@ jobs:
         run: .venv-ruleset/bin/python -m pip install --upgrade pip pytest 'ruff==0.15.11'
 
       - name: Run CI contract tests
-        run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py -q
+        run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py tests/tools/test_privileged_ruleset_audit.py -q
 
       - name: Compare live ruleset to checked-in snapshot
         env:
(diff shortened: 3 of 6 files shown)


tests/tools/test_privileged_ruleset_audit.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+    )
+
+
+def test_privileged_ruleset_audit_shares_ruleset_gate_semantics() -> None:
+    module = _load_audit_module()
+
+    assert (
+        module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
+        == module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
```
zero-bang: This assertion compares `module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS` to itself — it is always `True` and tests nothing.

```python
assert (
    module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
    == module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS  # both sides identical
)
```

To actually verify the audit tool shares the gate's constants, compare against the expected literal (consistent with how the other assertions in this test are written):

```python
assert (
    module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
    == frozenset({'name', 'target', 'enforcement', 'conditions', 'rules'})
)
```

This way the test will catch a regression if someone were to decouple the two modules or change the field set without updating both surfaces.

zero-bang (changes requested):
Solid addition — the privileged post-merge audit is well-scoped (main-only, `workflow_dispatch`), shares normalization with `ruleset_gate.py` rather than reimplementing, and the test coverage hits all the key behaviors (pass, drift, missing field, snapshot-on-failure). The workflow contract test correctly asserts `pull_request:` is absent, which is the critical security property for the privileged token.

One blocking finding: a tautological assertion in the "shares semantics" test.

mikkokotila:
## PR review against issue #163

Implementation tracks the slice spec faithfully. The shared-semantics design via `importlib.util.spec_from_file_location('ruleset_gate', TOOLS_DIR / 'ruleset_gate.py')` is exactly the pattern requested in the issue's third-pass review — the audit imports `REQUIRED_TOP_LEVEL_FIELDS`, `OPTIONAL_LIVE_TOP_LEVEL_FIELDS`, `IGNORED_LIVE_FIELDS`, `normalize_snapshot_ruleset`, `load_json_file`, `load_live_ruleset`, and `fail` from `tools/ruleset_gate.py` instead of redefining them.

The fail-loud-on-missing-`bypass_actors` contract is precisely what the slice promised:

```python
missing_optional = shared_ruleset_gate.OPTIONAL_LIVE_TOP_LEVEL_FIELDS - live_fields
if missing_optional:
    raise SystemExit(
        shared_ruleset_gate.fail(
            'privileged live ruleset missing required observable field(s): '
            f'{sorted(missing_optional)}',
            code=2,  # setup failure, not data failure
        )
    )
```

Code 2 (setup failure) is the right exit code — it signals "the privileged path could not run as designed" rather than "the data was wrong." The named test [`test_privileged_ruleset_audit_fails_loud_when_live_payload_omits_bypass_actors`](https://github.com/Vaquum/tdw-control-plane/blob/codex/privileged-ruleset-audit-main/tests/tools/test_privileged_ruleset_audit.py#L77-L96) pins the contract end-to-end: feeds the `_without_bypass_actors` fixture, asserts code 2, asserts the precise stderr message, asserts the snapshot is written.

The workflow's security boundary is also well-pinned by `test_audit_main_ruleset_workflow_contract`: `push:` and `workflow_dispatch:` are present, `pull_request:` is asserted **absent**. So a future PR that adds a `pull_request:` trigger to `audit_main_ruleset.yml` would fail the contract test — the credential boundary is mechanically enforced.

### Status — `pr_checks_version` failed on transient infrastructure

```
error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500
fatal: expected flush after ref listing
##[error]Process completed with exit code 128.
```

This is GitHub's git server returning HTTP 500 on the `git fetch origin main --depth=1` step in the `pr_checks_version` workflow. The version gate itself never ran. The PR's content is correct: pyproject bumps `1.6.2 → 1.6.3` (patch — consistent with `ci:` title), and CHANGELOG has a v1.6.3 section with three bullet points. Re-running the workflow should resolve this without any code change.

### Finding 1 — `test_privileged_ruleset_audit_shares_ruleset_gate_semantics` has a self-comparison bug

[Lines 102-105 of the test file](https://github.com/Vaquum/tdw-control-plane/blob/codex/privileged-ruleset-audit-main/tests/tools/test_privileged_ruleset_audit.py#L102-L105):

```python
assert (
    module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
    == module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
)
```

Both sides of the equality are the same expression — this is a no-op tautology. It does not verify the test's stated claim *"audit shares ruleset_gate semantics"* because there is no comparison to a separately-loaded `ruleset_gate` reference. The test passes trivially even if the audit tool ever stops sharing with `ruleset_gate.py`.

Fix: compare the audit's imported reference to a separately-loaded module. Something like:

```python
import importlib.util
spec = importlib.util.spec_from_file_location(
    'ruleset_gate_independent', REPO_ROOT / 'tools/ruleset_gate.py'
)
ruleset_gate = importlib.util.module_from_spec(spec)
spec.loader.exec_module(ruleset_gate)

assert (
    module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
    == ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
)
```

Or, since the audit's `shared_ruleset_gate` IS supposed to be `ruleset_gate.py` loaded via importlib, simpler:

```python
assert module.shared_ruleset_gate.__name__ == 'ruleset_gate'
assert module.shared_ruleset_gate.__file__ == str(REPO_ROOT / 'tools/ruleset_gate.py')
```

The other three assertions in the same test ARE substantive (`OPTIONAL_LIVE_TOP_LEVEL_FIELDS == frozenset({'bypass_actors'})`, `'bypass_actors' in SNAPSHOT_TOP_LEVEL_FIELDS`, `'bypass_actors' not in IGNORED_LIVE_FIELDS`). Just this one is broken.

### Finding 2 — pytest installed but unused in the audit workflow

[`audit_main_ruleset.yml` L29](https://github.com/Vaquum/tdw-control-plane/blob/codex/privileged-ruleset-audit-main/.github/workflows/audit_main_ruleset.yml#L29):

```yaml
- name: Install audit dependencies
  run: .venv-ruleset-audit/bin/python -m pip install --upgrade pip pytest
```

The audit step that follows runs `tools/privileged_ruleset_audit.py` directly — it is a standalone script with argparse, not pytest-driven. The audit tool's only imports are stdlib (`argparse`, `importlib.util`, `json`, `sys`, `types`, `pathlib`, `typing`) plus `ruleset_gate.py` loaded via importlib (also stdlib-only). Pytest is dead weight here.

Probably a copy-paste from `pr_checks_ruleset.yml`'s install step. Removing it shaves a few seconds off every audit run and reduces the surface area of "what's installed in the workflow that has the privileged token."

### Finding 3 — `--ruleset-id` parsed as `int` then cast back to `str`

```python
parser.add_argument('--ruleset-id', required=True, type=int)
...
return run_audit(
    ruleset_id=str(args.ruleset_id),  # cast back to string
    ...
)
```

`run_audit`'s parameter annotation is `ruleset_id: str`. The `type=int` does light validation (rejects non-integer strings) but the value flows through as `str` regardless. Either:

- declare `ruleset_id: int` end-to-end (cleaner — types match)
- declare `ruleset_id: str` and drop the `type=int` parser arg (simpler — one fewer cast)

Pure code smell, no functional impact.

### Operator note for closeout

The slice's Closeout requires `audit_main_ruleset:` as a CI run. That workflow only fires on `push: [main]` or `workflow_dispatch:`, so the evidence run id will come from the **first push to main after this PR merges** (i.e., the merge commit itself triggers the audit). Worth a one-line note in the Closeout fields when filling them in: *"audit_main_ruleset run id is from the post-merge run on commit `<sha>`, not from a PR-time run."* Mechanical, but easy to miss when filling closeout fields by hand.

### Verdict

After Finding 1 (self-comparison bug) is fixed and `pr_checks_version` is re-run, this PR is ready to merge. Findings 2 and 3 are cleanup that can land in this PR or a sibling. The substantive design — shared semantics via importlib, exit-2 on missing `bypass_actors`, workflow security boundary mechanically enforced — all matches what the slice promised.

zero-bang (approved):

mikkokotila:
## Second pass — all three findings closed cleanly

The new commit `d6f1585 ci: tighten privileged ruleset audit contract` addresses every finding from my prior review:

**Finding 1 (self-comparison bug)** — closed. New `_load_ruleset_gate_module()` helper loads `tools/ruleset_gate.py` independently, and `test_privileged_ruleset_audit_shares_ruleset_gate_semantics` now does four real cross-module comparisons:

```python
assert (
    module.shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
    == ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS
)
assert (
    module.shared_ruleset_gate.OPTIONAL_LIVE_TOP_LEVEL_FIELDS
    == ruleset_gate.OPTIONAL_LIVE_TOP_LEVEL_FIELDS
)
assert module.shared_ruleset_gate.SNAPSHOT_TOP_LEVEL_FIELDS == ruleset_gate.SNAPSHOT_TOP_LEVEL_FIELDS
assert module.shared_ruleset_gate.IGNORED_LIVE_FIELDS == ruleset_gate.IGNORED_LIVE_FIELDS
```

The test now actually catches divergence between the two module references — if a future PR adds a field to `ruleset_gate.py` but doesn't propagate to the audit's import path, this test fails.

**Finding 2 (pytest dead dependency)** — closed. Workflow install step is now `pip install --upgrade pip` only. Reduces what's installed in the workflow that holds `RULESET_AUDIT_TOKEN`.

**Finding 3 (`ruleset_id` int↔str round-trip)** — closed. `run_audit`'s signature is now `ruleset_id: int`; the conversion to `str` happens once at the boundary where `load_live_ruleset` (which expects `str`) is called. End-to-end types are consistent; the round-trip cast is gone.

**Bonus tightening I didn't flag:** the audit tool now has its own `fail()` function with `'privileged_ruleset_audit:'` prefix instead of borrowing `shared_ruleset_gate.fail()` (which prefixes with `'ruleset_gate:'`). Audit failures are now diagnostically distinguishable from ruleset_gate failures even though they share comparison semantics. Small but right call.

### Status

All 10 unique required checks `success` on head `d6f1585`. `mergeable_state: clean`. PR is ready to merge.

The `pr_checks_version` transient infrastructure failure from the previous head ran clean this time, confirming it was indeed git-server flakiness, not a content issue.

### Verdict

PR is ready to merge. The slice has gone through three honest revisions (initial → tighten privileged ruleset audit contract → operator-facing diagnostics) and the result is a clean post-merge audit path that closes the bypass_actors silent-gap from #160 without exposing the privileged credential to PR-time code. The shared-semantics design via importlib + the now-real cross-module test means the two ruleset comparison paths can't drift silently.

Operator reminder for closeout: the `audit_main_ruleset` run id required by #163's Closeout will come from the post-merge run on the merge SHA (the `push: [main]` trigger), not from any PR-time run.

.github/workflows/audit_main_ruleset.yml
```diff
@@ -0,0 +1,46 @@
+name: audit_main_ruleset
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
```
mikkokotila, line 6: `workflow_dispatch` breaks the claimed main-only boundary for this privileged-token workflow. GitHub documents that manual runs can target another branch or tag (`Run workflow` branch selector / `gh workflow run --ref`), so once this workflow exists on default branch, a later branch can modify it and still execute it with `RULESET_AUDIT_TOKEN`. Exact fix: either drop `workflow_dispatch`, or add a job-level hard guard like `if: github.ref == refs/heads/main` so manual runs on non-`main` refs fail before secret use.

bit-mis:
### Intentional Simplicity

**P1**
- Hand-rolled `importlib.util.spec_from_file_location` dance to load a sibling tool — `tools/privileged_ruleset_audit.py:14` — `tools/` already gets prepended to `sys.path` whenever the script runs as `python tools/privileged_ruleset_audit.py` (the workflow's invocation pattern), so `import ruleset_gate as shared_ruleset_gate` replaces the entire 10-line `_load_shared_ruleset_gate_module` helper. No other file in `tools/` reaches for `importlib` to pull in a sibling; this PR is introducing a new, more complex pattern for no functional gain.
- `run_audit` advertises `-> int` but raises `SystemExit` on validation failure — `tools/privileged_ruleset_audit.py:102` — `normalize_privileged_live_ruleset` does `raise SystemExit(fail(...))` for unexpected fields / missing `bypass_actors`, then `run_audit` catches `SystemExit`, writes the snapshot, and re-raises. The signature lies, and tests are forced to wrap calls in `try/except SystemExit` (`tests/tools/test_privileged_ruleset_audit.py:39`). A flat early-return path — validate → on error write snapshot and `return code`; on drift write snapshot and `return 1`; else `return 0` — removes the exception-as-control-flow, makes the return contract honest, and simplifies the test harness, all without changing CLI behaviour (`sys.exit(main())` already handles process-exit codes).

bit-mis:
### No Saga Ornamentation

**P1**
- Tautological "shared semantics" test — `tests/tools/test_privileged_ruleset_audit.py:205` — `test_privileged_ruleset_audit_shares_ruleset_gate_semantics` loads `tools/ruleset_gate.py` twice (once via the audit module's `shared_ruleset_gate`, once directly) and asserts the constants are equal. Both loads point at the same source file, so the equality holds by construction regardless of what the audit tool actually does. It dresses up the "must share semantics with `ruleset_gate.py`" risk from the spec but verifies nothing the rest of the suite doesn't already exercise. Drop it or replace with a test that asserts the audit's normalization actually delegates (e.g. patch `shared_ruleset_gate.REQUIRED_TOP_LEVEL_FIELDS` and observe the audit's behavior change).
- Decorative sibling-import helper — `tools/privileged_ruleset_audit.py:14` — `_load_shared_ruleset_gate_module` is a 9-line `importlib.util.spec_from_file_location` ceremony to import a file sitting next to it. No other tool in `tools/` does this; they all use stdlib only. When invoked as `python tools/privileged_ruleset_audit.py`, the script's directory is already on `sys.path[0]`, so `import ruleset_gate as shared_ruleset_gate` is sufficient. The only reason for the dynamic loader is the test's choice to load the audit module via `spec_from_file_location`; that should be fixed in the test (one-line `sys.path.insert`), not by exporting test-driven complexity into the production tool.

.github/workflows/audit_main_ruleset.yml
bit-mis: Not addressed. The workflow at .github/workflows/audit_main_ruleset.yml still declares `workflow_dispatch:` on line 6 with no job-level `if: github.ref == 'refs/heads/main'` guard, so the main-only boundary for the privileged `RULESET_AUDIT_TOKEN` is still breakable: once this file lives on the default branch, a later branch can modify it and trigger a manual run via `gh workflow run --ref` and still receive the secret. Please apply one of the suggested fixes — drop `workflow_dispatch`, or add a hard ref guard on the job before the secret-using step.

mikkokotila: Fixed in `c21482c`: the `audit_main_ruleset` job now has `if: github.ref == 'refs/heads/main'`, so manual dispatch on any non-main ref skips before checkout or secret use. The workflow contract test now asserts that guard.

mikkokotila:
Addressed the remaining merge-readiness items in `c21482c`:

- merged current `main` into the branch and bumped `1.6.6 -> 1.6.7`
- added a job-level `github.ref == 'refs/heads/main'` guard to the privileged audit workflow
- replaced the production `importlib` sibling-loader with direct `import ruleset_gate as shared_ruleset_gate`
- changed `run_audit` to return exit codes directly on validation failure instead of raising `SystemExit` internally
- removed the redundant shared-semantics self-load test and kept the behavioral contract tests

Local gates listed in the PR body are passing.

zero-bang (approved):

PR #188, opened 2026-04-30
fix(huggingface): publish spot klines from Origo projection

### LINKED ISSUE #187: fix(huggingface): publish spot klines from Origo projection
## Slice ID

S16

## Slice Title

fix(huggingface): publish spot klines from Origo projection

## Routing

Parent issue: none
Depends on:
- #185
Unlocks:
- (none)

## Minimum Viable Capability

>
> Accepted:
>
> Rejected:
> - *Strict CQRS separation.*

- `python3 -m pytest tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py::test_publish_sensor_targets_origo_spot_trades_materialization -q` exits `0`
- `python3 -m pytest tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py::test_publish_snapshot_reads_origo_spot_trades_with_shared_query -q` exits `0`

## Current State

>
> Accepted:
>
> Rejected:

- `git show origin/main:tdw_control_plane/definitions.py | rg -n 'asset_key=AssetKey\("insert_daily_binance_trades_to_tdw"\)' | wc -l | tr -d ' '` returns `1`
- `git show origin/main:tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py | rg -n 'tdw\.binance_trades_complete|table_name="binance_trades_complete"|from tdw_control_plane\.query import get_binance_spot_klines' | wc -l | tr -d ' '` returns `5`
- `git show origin/main:tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py | rg -n 'origo\.binance_daily_spot_trades|refresh_binance_spot_klines_origo'; echo EXIT:$?` returns `EXIT:1`
- `git show origin/main:tdw_control_plane/query/get_binance_spot_klines.py | rg -n 'database_name|FROM \{database_name\}'; echo EXIT:$?` returns `EXIT:1`
- `git show origin/main:pyproject.toml | python3 -c "import sys, tomllib; print(tomllib.loads(sys.stdin.read())['project']['version'])"` returns `1.6.5`

## Target State

> predicate.
>
> Accepted:

- `python3 -m pytest tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py::test_publish_sensor_targets_origo_spot_trades_materialization -q` exits `0`
- `python3 -m pytest tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py::test_publish_snapshot_reads_origo_spot_trades_with_shared_query -q` exits `0`
- `git grep -n 'asset_key=AssetKey("insert_daily_binance_spot_trades_to_origo")' -- tdw_control_plane/definitions.py | wc -l | tr -d ' '` returns `1`
- `grep -q 'origo.binance_daily_spot_trades' tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py; echo $?` returns `0`
- `git grep -n 'database_name="origo"' -- tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py | wc -l | tr -d ' '` returns `1`
- `git grep -n 'database_name: str = "tdw"' -- tdw_control_plane/query/get_binance_spot_klines.py | wc -l | tr -d ' '` returns `1`
- `git grep -n 'binance_trades_complete' -- tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py; echo EXIT:$?` returns `EXIT:1`
- `python3 -c "import tomllib, pathlib; print(tomllib.loads(pathlib.Path('pyproject.toml').read_text())['project']['version'])"` returns `1.6.6`
- `grep -qE '^# v1\.6\.6\b' CHANGELOG.md; echo $?` returns `0`

## Surfaces

- `tdw_control_plane/query/get_binance_spot_klines.py`
- `pyproject.toml`
- `CHANGELOG.md`

Added:
- `tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py`

Removed:
- (none)

## Signature Diff

### tdw_control_plane/definitions.py::publish_binance_spot_klines_to_huggingface_sensor
Before:
```python
@asset_sensor(
    asset_key=AssetKey("insert_daily_binance_trades_to_tdw"),
    job=publish_binance_spot_klines_to_huggingface_job,
)
def publish_binance_spot_klines_to_huggingface_sensor(context, asset_event):
```
After:
```python
@asset_sensor(
    asset_key=AssetKey("insert_daily_binance_spot_trades_to_origo"),
    job=publish_binance_spot_klines_to_huggingface_job,
)
def publish_binance_spot_klines_to_huggingface_sensor(context, asset_event):
```

### tdw_control_plane/query/get_binance_spot_klines.py::get_binance_spot_klines
Before:
```python
def get_binance_spot_klines(
    n_rows: int | None = None,
    kline_size: int = 1,
    start_date_limit: str | None = None,
    end_date_limit: str | None = None,
    show_summary: bool = False,
    table_name: str = "binance_trades_complete",
    include_quantiles: bool = True,
) -> pl.DataFrame:
```
After:
```python
def get_binance_spot_klines(
    n_rows: int | None = None,
    kline_size: int = 1,
    start_date_limit: str | None = None,
    end_date_limit: str | None = None,
    show_summary: bool = False,
    table_name: str = "binance_trades_complete",
    database_name: str = "tdw",
    include_quantiles: bool = True,
) -> pl.DataFrame:
```

### tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py::source import
Before:
```python
from tdw_control_plane.assets.daily_trades_to_tdw import daily_partitions
from tdw_control_plane.query import get_binance_spot_klines
```
After:
```python
from tdw_control_plane.assets.daily_trades_to_origo import daily_partitions
from tdw_control_plane.query import get_binance_spot_klines
```

### tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py::source query
Before:
```python
data = get_binance_spot_klines(
    kline_size=60,
    start_date_limit=EXPORT_START_DATE,
    end_date_limit=export_end_exclusive,
    table_name="binance_trades_complete",
    include_quantiles=False,
)
```
After:
```python
data = get_binance_spot_klines(
    kline_size=60,
    start_date_limit=EXPORT_START_DATE,
    end_date_limit=export_end_exclusive,
    table_name="binance_daily_spot_trades",
    database_name="origo",
    include_quantiles=False,
)
```

### tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py::_build_dataset_card source line
Before:
```markdown
This dataset is exported daily from `tdw.binance_trades_complete` using TDW's `get_binance_spot_klines` query at 1-minute resolution.
```
After:
```markdown
This dataset is exported daily from `origo.binance_daily_spot_trades` using `get_binance_spot_klines` at 1-minute resolution.
```

Before:
```toml
version = "1.6.5"
```
After:
```toml
version = "1.6.6"
```

Before:
```markdown
# v1.6.5 on April 28, 2026
```
After:
```markdown
# v1.6.6 on April 30, 2026
```

## Tests That Prove the MVC

|---|---|---|
| HF publish sensor targets the Origo source-native spot trades materialization | `tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py` | `test_publish_sensor_targets_origo_spot_trades_materialization` |
| HF publish snapshot uses the shared spot-kline query against Origo source-native trades and publishes the requested partition | `tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py` | `test_publish_snapshot_reads_origo_spot_trades_with_shared_query` |

## Regression Gate

>
> Accepted:
>   on non-zero exit.
>
> Rejected:

Workflow file:
Job name:
- `pr_checks_tests`

## Out of Scope

- `tdw_control_plane/assets/daily_trades_to_tdw.py`
- `tdw_control_plane/assets/refresh_binance_spot_klines_origo.py`
- `tdw_control_plane/assets/refresh_aligned_1m_exchange_from_binance_spot_origo.py`
- `tests/fixtures/**`
- `.github/workflows/**`
- `.github/rulesets/**`

## Risks and Open Questions

Risks:
- This slice intentionally does not repair or extend the old TDW daily overlay schedule. It removes the HF dependency on that path by switching the trigger and export source to Origo.
- The publisher should continue to omit `median` and `iqr` from the exported Parquet snapshot to preserve the existing HF dataset column contract.

Open questions:
- None.

## Closeout

> bot verifies:

Merge SHA:
Merged PR number:
- `PR Checks CodeQL (python)`:
- `pr_checks_typing`:
- `pr_checks_cc`:
- `pr_checks_slice`:
- `pr_checks_fail_loud`:
- `pr_checks_version`:
- `pr_checks_ruleset`:
- `pr_checks_lint`:
- `pr_checks_tests`:

## Author Checks

## Appendix: Investigation Report

The publish job itself is not firing because its sensor has no new upstream event to consume.

The sensor is `RUNNING`, but every tick is skipped with:

```text
No new materialization events found for asset key AssetKey(['insert_daily_binance_trades_to_tdw'])
```

That is the key. The HF sensor is wired to the old TDW daily asset:

```python
@asset_sensor(
    asset_key=AssetKey("insert_daily_binance_trades_to_tdw"),
    job=publish_binance_spot_klines_to_huggingface_job,
)
```

But that TDW daily asset has not materialized anything after `2026-04-14`.

Current Dagster asset truth:

```text
insert_daily_binance_trades_to_tdw:
  materialized only 2026-04-01 through 2026-04-14
  2026-04-15 through 2026-04-29 are MISSING

publish_binance_spot_klines_to_huggingface:
  latest materialized partition: 2026-04-14
  latest sensor-launched run: 2026-04-27 11:13 UTC
```

Why the TDW daily asset stopped:

`daily_tdw_pipeline_schedule` is running, but it has been skipping every tick since `2026-04-16` with:

```text
Daily overlay gap is larger than the automated backfill threshold; trigger a manual backfill.
```

The schedule computes its gap from `tdw.binance_trades`, whose latest monthly day is `2026-03-31`, so it always starts at `2026-04-01`. On `2026-04-30`, it sees:

```text
start_day = 2026-04-01
end_date  = 2026-04-29
gap       = 29 days
limit     = 14 days
```

So it refuses to request any more daily TDW overlay runs, even though `2026-04-01..14` already exist.

Table truth matches that:

```text
tdw.binance_trades              max day 2026-03-31
tdw.binance_daily_trades        max day 2026-04-14
tdw.binance_trades_complete     max day 2026-04-14
origo.binance_spot_klines       max day 2026-04-29
origo aligned/binance_spot      max day 2026-04-29
```

The remote HF `latest.json` also confirms the published dataset is stale:

```text
export_end_date: 2026-04-14
generated_at_utc: 2026-04-27T11:13:49Z
```

So the root problem is not HF auth and not a stopped sensor. It is that the HF automation is still tied to the old TDW daily overlay path, while the fresh data is now in Origo. The old TDW path froze because its schedule's 14-day guard is computed from the monthly table boundary, not from the latest existing daily overlay day.

llun-ved:
Closes #187

Implements the slice as specified in #187.


bit-mis:
### Intentional Simplicity

**P1**
- `_get_origo_spot_klines` reinvents this repo's standard ClickHouse client construction — `tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py:59` — every other origo-touching asset imports `_get_clickhouse_settings` / `_make_clickhouse_client` from `create_origo_database` (e.g. `refresh_binance_spot_klines_origo.py:11`, `daily_trades_to_origo.py:17`, `refresh_aligned_1m_exchange_from_binance_spot_origo.py:12`), so duplicating the env-var reads, defaults, and `ClickhouseClient(...)` call here is accidental complexity that diverges from the established convention. The pandas detour (`settings={"use_numpy": True}` + `query_dataframe` + `pl.from_pandas`) is also unnecessary: the new test on the same slice does the equivalent read with `client.execute(...)` rows + `pl.DataFrame(rows, schema=_ORIGO_SPOT_KLINE_COLUMNS, orient='row')` (`tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py:271`), which is the simpler, polars-only path the slice actually needs.

bit-mis:
### Concrete Value for the User

**P1**
- Silent regression in the published HF parquet schema. The previous publisher path (`tdw_control_plane/query/get_binance_spot_klines.py:143-155`) explicitly normalised the wire format consumers actually download — `datetime` cast to `pl.Datetime("ms", time_zone="UTC")` and `mean / std / volume / liquidity_sum / maker_liquidity` rounded to 5/6/9/1/1 decimals. The new path returns whatever `clickhouse_driver.query_dataframe(..., settings={"use_numpy": True})` + `pl.from_pandas(...)` happens to emit (tz-naive `Datetime("ns")` and unrounded Float64), and `data.write_parquet(...)` ships that directly. The added test passes `check_dtypes=False` and compares against the same origo source, so the divergence from the previously published contract is not detected. The spec explicitly calls out "preserve the existing HF dataset column contract" — column presence is preserved, but the on-disk dtype/precision the user actually consumes is not. — `tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py:75-112` — the user-observable artefact (the daily parquet on Hugging Face) silently changes shape on the same merge that claims only to switch the source projection.

bit-mis (approved):
bit-mis: all green; no unresolved P0/P1 from facet review or audit. Approving.

tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
```diff
@@ -32,10 +56,50 @@ def _get_huggingface_dataset_repo_id() -> str:
...
+        raise RuntimeError(
+            "CLICKHOUSE_PASSWORD environment variable must be set before reading Origo klines."
+        )
+
+    host = os.environ.get("CLICKHOUSE_HOST", "clickhouse")
+    port = int(os.environ.get("CLICKHOUSE_PORT", "9000"))
+    user = os.environ.get("CLICKHOUSE_USER", "default")
+    database = os.environ.get("CLICKHOUSE_DATABASE", "origo")
+
+    client = ClickhouseClient(
```
mikkokotila: This hand-rolls a new ClickHouse client path inside the publisher even though the Origo assets already use `_get_clickhouse_settings()` / `_make_clickhouse_client()` from `create_origo_database`. The slice only needs to read the existing `origo.binance_spot_klines` projection, so this should be a tiny query using the established Origo client convention, ideally without the extra pandas `query_dataframe` hop.

tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
```diff
@@ -32,10 +56,50 @@ def _get_huggingface_dataset_repo_id() -> str:
...
+            WHERE datetime >= toDateTime(%(start_dt)s)
+              AND datetime <  toDateTime(%(end_dt)s)
+            ORDER BY datetime
+            """,
+            {"start_dt": start_date_limit, "end_dt": end_date_limit},
+        )
+    finally:
+        client.disconnect()
+
+    return pl.from_pandas(pandas_df)
```
mikkokotila: Returning raw `pl.from_pandas(pandas_df)` changes the HF Parquet contract. The old `get_binance_spot_klines` path normalized `datetime` to `pl.Datetime("ms", time_zone="UTC")` and rounded `mean`, `std`, `volume`, `liquidity_sum`, and `maker_liquidity` before `write_parquet`; this path writes ClickHouse/pandas defaults directly. The new test compares against the same raw Origo source with `check_dtypes=False`, so it will not catch the user-visible dtype/precision regression.

mikkokotila (changes requested):
I agree with the scope concern here. The slice should be a narrow reroute: trigger from the Origo materialization and read the already-materialized Origo kline projection while preserving the existing HF artifact contract. As written, the PR adds a new bespoke ClickHouse/pandas read path and does not preserve the previous Parquet dtype/precision normalization, so I would not merge this shape.

mikkokotila:
To make it clear, this PR is only about rerouting the publish trigger, and only that.

bit-mis (approved):
bit-mis: all green; no unresolved P0/P1 from facet review or audit. Approving.

bit-mis (approved):
bit-mis: all green; no unresolved P0/P1 from facet review or audit. Approving.

tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
```diff
@@ -5,15 +5,39 @@
...
+    "volume",
+    "maker_ratio",
+    "no_of_trades",
+    "open_liquidity",
+    "high_liquidity",
+    "low_liquidity",
+    "close_liquidity",
+    "liquidity_sum",
+    "maker_volume",
+    "maker_liquidity",
```
mikkokotila, line 38: Why do we add this slice is about rerouting and nothing else. The fix has been identified in the spec, and this change seems to have nothing to do with it.

tdw_control_plane/assets/publish_binance_spot_klines_to_huggingface.py
```diff
@@ -32,10 +56,46 @@ def _get_huggingface_dataset_repo_id() -> str:
...
+        pl.col("mean").round(5),
+        pl.col("std").round(6),
+        pl.col("volume").round(9),
+        pl.col("liquidity_sum").round(1),
+        pl.col("maker_liquidity").round(1),
+    ])
+
+    return polars_df
+
+
```
mikkokotila, line 94: Why do we add this slice is about rerouting and nothing else. The fix has been identified in the spec, and this change seems to have nothing to do with it. The asset was working just fine, it just needs to be routed to a different table, that's all.

mikkokotila (changes requested):

PR #192, opened 2026-05-14
feat(origo): add binance spot depth20 data source

mikkokotila:
Closes #191.

## Summary
- Add source-native Binance spot depth20 snapshots table and history sync.
- Add per-source 1m depth20 projection with `datetime`, `source_timestamp_ms`, and the five agreed scalar book columns.
- Register the every-minute UTC Dagster refresh job and pass API credentials through repo secrets in both deploy validation and deploy runtime.

## Checks
- PASS: compileall tdw_control_plane
- PASS: targeted pyright on new depth20 assets
- PASS: typing_gate with pyright summary 1193 errors <= budget 1213
- PASS: fail_loud_gate
- PASS: ruff check tools tests/tools
- PASS: docker compose -f docker-compose.deploy.yml config --quiet
- PASS: git diff --check
- BLOCKED locally: pytest tests/origo_source_native needs Docker; daemon socket is unavailable on this machine.


zero-bang (approved):
Thorough review complete. This is a clean, well-structured PR.

**What I verified:**
- Snapshot ingestion pipeline: history API download → parse NDJSON → delete-then-insert into ClickHouse with `mutations_sync: 2`. Idempotency is sound.
- 1m projection: picks the last snapshot in the minute window (`ORDER BY datetime DESC LIMIT 1`), computes 5 scalar book columns. ClickHouse's column-alias-forwarding in SELECT makes the `book_mid_price` → `book_spread_bps` reference work correctly.
- Schedule: every-minute cron with `_last_completed_minute` correctly subtracting 1 minute from scheduled time. Run key includes the ISO timestamp for dedup.
- Deployment: env vars plumbed through CI validation (test values), deploy secrets, and docker-compose (both webserver and daemon). All use `${VAR:?}` required syntax.
- Tests: comprehensive coverage — schema contracts, fixture checksum integrity, data round-trip with independent Python verification via `pytest.approx`, idempotency, and Dagster job/schedule registration assertions.
- Fixture timestamps: verified both rows (1778678580141ms, 1778678639138ms) fall within the 13:23 UTC minute, and `int(minute_start.timestamp())` correctly round-trips to 1778678580.
- Version bump 1.6.7 → 1.7.0 is appropriate for a new feature. CHANGELOG entry is accurate.

**Forward-looking note:** If the history API is unavailable for consecutive minutes, those minutes will fail without automatic backfill — the schedule always targets `now - 1 minute`. Worth considering a catch-up mechanism if sustained gaps are operationally significant, but not blocking for this PR.

LGTM

tdw_control_plane/assets/sync_binance_spot_depth20_snapshots_to_origo.py
```diff
@@ -0,0 +1,164 @@
...
+
+def _delete_minute_rows(
+    client: ClickHouseClient,
+    database: str,
+    minute_start: datetime,
+) -> None:
+    minute_end = minute_start + timedelta(minutes=1)
+    client.execute(
+        f"""
+        ALTER TABLE {database}.{SNAPSHOTS_TABLE_NAME}
```
bit-mis: P2: This `ALTER TABLE ... DELETE` is driven by the new `* * * * *` `binance_spot_depth20_1m_schedule`, so it enqueues a row in `system.mutations` and forces a part rewrite ~1,440 times/day on `binance_spot_depth20_snapshots` (and another ~1,440/day on the 1m projection via the equivalent block in `refresh_binance_spot_depth20_1m_origo.py::_delete_minute_row`). The existing klines refresh uses the same delete-then-insert pattern, but at daily cadence — 60x less frequent — where mutations have time to settle and the `mutations_sync: 2` cost is amortized. At minute cadence on `MergeTree`, mutation backlog and part rewrites become an ongoing operational tax on every monthly partition the schedule touches. For minute-cadence idempotency, the established ClickHouse pattern is `ReplacingMergeTree(source_timestamp_ms)` (or another monotonically increasing version column) with `ORDER BY (datetime)` and insert-only semantics, or plain `MergeTree` with insert-only writes plus `argMax`/`FINAL` at read time. Either eliminates the per-minute mutation entirely.

bit-mis (changes requested):

zero-bang (approved):

bit-mis (approved):

zero-bang (approved):

zero-bang (approved):

bit-mis (approved):

PR #198, opened 2026-05-19
feat(origo): add binance spot dollar klines

mikkokotila:
## Summary
- Add `origo.binance_spot_dollar_klines` with the existing spot kline statistic footprint plus `dollar_bar_id`.
- Add the daily dollar-kline refresh asset and register it in the Binance spot Origo data-source job.
- Add Origo source-native tests, version bump, and changelog entry.

## Validation
- `ruff check tdw_control_plane/assets/create_binance_spot_dollar_klines_table_origo.py tdw_control_plane/assets/refresh_binance_spot_dollar_klines_origo.py tests/origo_source_native/test_origo_binance_spot_dollar_klines_template.py tools tests/tools`
- `python3 -m py_compile tdw_control_plane/assets/create_binance_spot_dollar_klines_table_origo.py tdw_control_plane/assets/refresh_binance_spot_dollar_klines_origo.py tests/origo_source_native/test_origo_binance_spot_dollar_klines_template.py`
- `pytest tests/origo_source_native/test_origo_binance_spot_dollar_klines_template.py -q --maxfail=1` blocked locally because Docker daemon is unavailable.

Closes #197

### DIFF
--- a/tdw_control_plane/assets/refresh_binance_spot_dollar_klines_origo.py
+++ b/tdw_control_plane/assets/refresh_binance_spot_dollar_klines_origo.py
@@ -112,7 +112,11 @@ def _insert_partition_rows(
         insert_daily_binance_spot_trades_to_origo,
     ],
     group_name='binance_data',
-    description='Refreshes the Binance spot dollar kline projection from source-native daily trades',
+    description=(
+        'Refreshes the daily-scoped Binance spot dollar kline projection from '
+        'source-native daily trades; dollar_bar_id resets each date and the final '
+        'daily bar may be below the dollar threshold.'
+    ),
 )
 def refresh_binance_spot_dollar_klines_origo(
     context: AssetExecutionContext,
--- a/tdw_control_plane/assets/refresh_binance_spot_dollar_klines_origo.py
+++ b/tdw_control_plane/assets/refresh_binance_spot_dollar_klines_origo.py
@@ -1,0 +1,143 @@
+from datetime import UTC, datetime, timedelta
+
+from clickhouse_driver import Client as ClickhouseClient
+from dagster import AssetExecutionContext, asset
+
+from .create_binance_spot_dollar_klines_table_origo import (
+    DOLLAR_KLINES_TABLE_NAME,
+    create_binance_spot_dollar_klines_table_origo,
+)
+from .create_binance_trades_table_origo import RAW_TABLE_NAME
+from .create_origo_database import _get_clickhouse_settings, _make_clickhouse_client
+from .daily_trades_to_origo import daily_partitions, insert_daily_binance_spot_trades_to_origo
+
+DOLLAR_KLINE_SIZE = 100_000.0
+
+
+def _partition_date_from_context(context: AssetExecutionContext) -> str:
+    partition_key = context.partition_key
+    if partition_key is not None:
+        return partition_key
+
+    target_date = datetime.now(UTC) - timedelta(days=1)
+    return target_date.strftime('%Y-%m-%d')
+
+
+def _delete_partition_rows(
+    client: ClickhouseClient,
+    database: str,
+    partition_date: str,
+) -> None:
+    client.execute(
+        f"""
+        ALTER TABLE {database}.{DOLLAR_KLINES_TABLE_NAME}
+        DELETE WHERE toDate(datetime) = toDate('{partition_date}')
+        """,
+        settings={'mutations_sync': 2},
+    )
+
+
+def _count_partition_rows(
+    client: ClickhouseClient,
+    database: str,
+    partition_date: str,
+) -> int:
+    result = client.execute(
+        f"""
+        SELECT count()
+        FROM {database}.{DOLLAR_KLINES_TABLE_NAME}
+        WHERE toDate(datetime) = toDate('{partition_date}')
+        """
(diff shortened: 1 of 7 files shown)


zero-bang (approved):
LGTM — thorough review, no issues found.

**Verified:**
- Dollar klines table schema correctly extends the spot klines footprint with `dollar_bar_id UInt64` and matching ORDER BY.
- Traced the window-function-based dollar bar assignment SQL through the fixture data (3 trades, threshold=40.0) — bar IDs 0, 1, 3 are correct (bar 2 skipped because no trade falls in the 80–120 cumulative quote range).
- All Dagster registration points covered: imports, job definition, data source job selection, assets list, jobs list.
- Test coverage is comprehensive: schema contract, row-level correctness, idempotency, and registration.
- Refresh follows the established delete-then-insert idempotency pattern with `mutations_sync: 2`.
- New asset correctly wired into `refresh_binance_spot_data_source_job`, so the existing `daily_binance_spot_pipeline_schedule` will pick it up automatically.

**Forward-looking note:** The dollar bar IDs are computed per-day (the running cumulative resets at day boundaries), which means the first and last bars of each day may be partial. This is consistent with how the regular klines scope to individual days, but worth noting if cross-day continuity ever becomes a requirement.

tdw_control_plane/assets/refresh_binance_spot_dollar_klines_origo.py
```diff
@@ -0,0 +1,143 @@
...
+                    *,
+                    greatest(
+                        sum(quote_quantity) OVER (
+                            ORDER BY datetime, trade_id
+                            ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
+                        ) - quote_quantity,
+                        0.0
+                    ) AS running_quote_before
+                FROM {database}.{RAW_TABLE_NAME}
+                WHERE toDate(datetime) = toDate('{partition_date}')
```
bit-mis, line 100: The window function `sum(quote_quantity) OVER (...)` runs only over rows kept by this `WHERE toDate(datetime) = toDate(...)` filter, so `running_quote_before` (and therefore `dollar_bar_id`) restarts at 0 for every partition. Two consequences are not surfaced anywhere in the asset description or table schema:

- `dollar_bar_id` is unique only within a single partition_date. A query like `WHERE dollar_bar_id = 0` returns one row per day, not globally — a composite `(toDate(datetime), dollar_bar_id)` is required to identify a bar.
- The last bar of each day is typically incomplete (does not reach `DOLLAR_KLINE_SIZE`), and the next day starts fresh at bar 0, so the dollar-volume series is *not* continuous across calendar boundaries — unlike canonical dollar bars.

If that is the intended methodology (daily-scoped dollar klines rather than continuous dollar bars), please call it out in the asset description and/or table-creation docstring so downstream consumers do not assume continuity. Otherwise, consider seeding `running_quote_before` from the prior partition tail (or emitting a globally unique `dollar_bar_id`) so bars stay continuous across days.

bit-mis (changes requested):

tdw_control_plane/assets/refresh_binance_spot_dollar_klines_origo.py
```diff
@@ -0,0 +1,143 @@
...
+            max(price * quantity) AS high_liquidity,
+            min(price * quantity) AS low_liquidity,
+            argMax(price * quantity, trade_id) AS close_liquidity,
+            sum(price * quantity) AS liquidity_sum,
+            sumKahan(is_buyer_maker * quantity) AS maker_volume,
+            sum(is_buyer_maker * price * quantity) AS maker_liquidity
+        FROM (
+            SELECT
+                *,
+                toUInt64(floor(running_quote_before / {DOLLAR_KLINE_SIZE})) AS dollar_bar_id
```
bit-mis, line 88: P1: `dollar_bar_id` is computed per partition — the window function above runs cumulative `sum(quote_quantity)` over a single day's rows only (the inner `WHERE toDate(datetime) = toDate('{partition_date}')` constrains its input), so the IDs reset to 0 at every day boundary and are NOT unique within the table. The de-facto unique key is `(toDate(datetime), dollar_bar_id)`, not `dollar_bar_id` alone.

The IDs are also non-contiguous within a day: any trade whose `quote_quantity` straddles multiple thresholds skips bar IDs — the fixture in `test_origo_binance_spot_dollar_klines_template.py` jumps from `dollar_bar_id=1` to `dollar_bar_id=3` for exactly this reason. Both properties diverge from the conventional notion of dollar bars (continuous across all trades).

The table description and the asset description say nothing about either property, and the schema has no column comment, so downstream consumers will reasonably assume `dollar_bar_id` is globally unique and/or contiguous. Please surface these semantics either in the asset description or as a `COMMENT` on the `dollar_bar_id` column in `create_binance_spot_dollar_klines_table_origo.py` so this is discoverable without reading the SQL.

bit-mis (changes requested):

zero-bang (approved):

bit-mis (approved):

zero-bang (approved):

bit-mis (approved):

bit-mis (approved):

zero-bang (approved):

PR #200, opened 2026-05-19
feat(huggingface): publish Binance spot 15m, 30m, and 2h klines

mikkokotila:
Closes #199

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v1.10.0 on May 19, 2026
+- Add Hugging Face publishers for BTCUSDT 15-minute, 30-minute, and 2-hour spot kline snapshots from Origo daily spot trades.
+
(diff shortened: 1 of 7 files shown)


zero-bang (approved):
LGTM — clean mechanical extension of the existing kline publisher pattern.

All three new assets (15m/900s, 30m/1800s, 2h/7200s) faithfully replicate the established 1h/4h structure: same imports, decorator shape, shared utility delegation, sensor wiring, and registration in all four definitions.py lists (imports, assets, sensors, jobs). Ascending cadence order is maintained throughout.

Run key prefixes are unique per cadence, repo IDs are distinct, and the test covers both sensor targeting and end-to-end publish flow with proper assertions on query parameters, file naming, commit messages, delete patterns, and README content. The consolidated test file for all three cadences is a good call over three near-identical files.

Version bump and changelog are correct.

PR #202, opened 2026-05-19
feat(origo): add binance spot volume klines

mikkokotila:
Closes #201

Summary:
- Add Origo Binance spot volume kline create/refresh assets.
- Register volume klines in the existing daily spot refresh job and create-table job.
- Add volume kline Origo source-native tests plus version/changelog trail.

Checks:
- Codex CLI issue review: APPROVED
- ruff check tdw_control_plane/assets/create_binance_spot_volume_klines_table_origo.py tdw_control_plane/assets/refresh_binance_spot_volume_klines_origo.py tests/origo_source_native/test_origo_binance_spot_volume_klines_template.py tools tests/tools: PASS
- python3 -m py_compile changed Python surfaces: PASS
- tools/fail_loud_gate.py: PASS
- tools/version_gate.py: PASS
- tools/typing_gate.py: PASS
- tools/slice_gate.py against issue #201: PASS
- pytest tests/origo_source_native -q --maxfail=1: BLOCKED locally, Docker daemon socket missing

### DIFF
--- a/tdw_control_plane/assets/create_binance_spot_volume_klines_table_origo.py
+++ b/tdw_control_plane/assets/create_binance_spot_volume_klines_table_origo.py
@@ -1,0 +1,179 @@
+import os
+from collections.abc import Mapping
+from dataclasses import dataclass
+from importlib import import_module
+from typing import Protocol, runtime_checkable
+
+from dagster import AssetExecutionContext, asset
+
+from .create_origo_database import create_origo_database
+
+_DEFAULT_CLICKHOUSE_HOST = 'clickhouse'
+_DEFAULT_CLICKHOUSE_PORT = 9000
+_DEFAULT_CLICKHOUSE_USER = 'default'
+VOLUME_KLINES_TABLE_NAME = 'binance_spot_volume_klines'
+
+
+@dataclass(frozen=True)
+class _ClickHouseSettings:
+    host: str
+    port: int
+    user: str
+    password: str
+    database: str
+
+
+@runtime_checkable
+class _ClickHouseClientLike(Protocol):
+    def execute(
+        self,
+        query: str,
+        params: object | None = None,
+        *,
+        settings: Mapping[str, object] | None = None,
+    ) -> list[tuple[object, ...]]:
+        raise NotImplementedError
+
+    def disconnect(self) -> None:
+        raise NotImplementedError
+
+
+def _require_env(name: str) -> str:
+    value = os.environ.get(name)
+    if not value:
+        raise RuntimeError(f'{name} environment variable must be set.')
+    return value
+
+
+def _get_clickhouse_port() -> int:
+    value = os.environ.get('CLICKHOUSE_PORT', str(_DEFAULT_CLICKHOUSE_PORT))
+    try:
+        return int(value)
+    except ValueError as exc:
+        raise RuntimeError('CLICKHOUSE_PORT environment variable must be an integer.') from exc
+
+
+def _get_clickhouse_settings() -> _ClickHouseSettings:
+    return _ClickHouseSettings(
+        host=os.environ.get('CLICKHOUSE_HOST', _DEFAULT_CLICKHOUSE_HOST),
+        port=_get_clickhouse_port(),
+        user=os.environ.get('CLICKHOUSE_USER', _DEFAULT_CLICKHOUSE_USER),
+        password=_require_env('CLICKHOUSE_PASSWORD'),
+        database=os.environ.get('CLICKHOUSE_DATABASE', 'origo'),
+    )
+
+
+def _make_clickhouse_client(settings: _ClickHouseSettings) -> _ClickHouseClientLike:
+    client_factory = getattr(import_module('clickhouse_driver'), 'Client')
+    client = client_factory(
+        host=settings.host,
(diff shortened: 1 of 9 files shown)


tdw_control_plane/assets/create_binance_spot_volume_klines_table_origo.py
```diff
@@ -0,0 +1,179 @@
...
+        user=settings.user,
+        password=settings.password,
+    )
+    if not isinstance(client, _ClickHouseClientLike):
+        raise TypeError('clickhouse_driver.Client does not satisfy the ClickHouse client contract.')
+    return client
+
+
+def _count_result(result: list[tuple[object, ...]]) -> int:
+    count_value = result[0][0]
```
zero-bang: Lines 1-80 redefine `_ClickHouseSettings`, `_require_env`, `_get_clickhouse_port`, `_get_clickhouse_settings`, `_make_clickhouse_client`, `_database_exists`, and `_table_exists` — all of which already exist as shared exports in `create_origo_database.py` and `create_binance_trades_table_origo.py`.

The dollar-klines create file (`create_binance_spot_dollar_klines_table_origo.py`, 81 lines) does this:
```python
from .create_binance_trades_table_origo import _database_exists, _table_exists
from .create_origo_database import (
    ClickHouseSettings,
    _get_clickhouse_settings,
    _make_clickhouse_client,
    create_origo_database,
)
```

Following that pattern would drop this file from 179 → ~80 lines and keep ClickHouse connection logic in one place. The `import_module` + `_ClickHouseClientLike` Protocol approach introduced here isn't used by the 7+ other asset files that share the common helpers.

tdw_control_plane/assets/refresh_binance_spot_volume_klines_origo.py
```diff
@@ -0,0 +1,219 @@
...
+    return _ClickHouseSettings(
+        host=os.environ.get('CLICKHOUSE_HOST', _DEFAULT_CLICKHOUSE_HOST),
+        port=_get_clickhouse_port(),
+        user=os.environ.get('CLICKHOUSE_USER', _DEFAULT_CLICKHOUSE_USER),
+        password=_require_env('CLICKHOUSE_PASSWORD'),
+        database=os.environ.get('CLICKHOUSE_DATABASE', 'origo'),
+    )
+
+
+def _make_clickhouse_client(settings: _ClickHouseSettings) -> _ClickHouseClientLike:
```
zero-bang: Same duplication issue as the create file: lines 1-72 re-implement settings, client construction, and env helpers that the dollar-klines refresh imports from shared modules in two lines:
```python
from .create_origo_database import _get_clickhouse_settings, _make_clickhouse_client
```
Following that pattern would drop this file from 219 → ~150 lines.

zero-bang (changes requested):
SQL logic, column schemas, definitions.py registrations, test coverage, and CHANGELOG/version are all correct. The volume bar calculation correctly mirrors the dollar bar pattern with `sum(quantity)` instead of `sum(quote_quantity)`, and the idempotent delete-then-insert refresh is sound.

One blocking concern: both new asset files inline ~170 lines of shared infrastructure that already exists in `create_origo_database.py` and `create_binance_trades_table_origo.py`. Seven existing asset files import these shared helpers; only one existing file (depth20 snapshots) inlines them. Following the dominant pattern (and the direct dollar-klines precedent) would cut ~170 lines and keep maintenance centralized.

zero-bang (approved):

tdw_control_plane/assets/create_origo_database.py
```diff
@@ -52,6 +78,17 @@ def _make_clickhouse_client(settings: ClickHouseSettings) -> ClickhouseClient:
...
 
 
+def get_clickhouse_settings() -> ClickHouseSettings:
+    return _get_clickhouse_settings()
+
+
+def make_clickhouse_client(settings: ClickHouseSettings) -> ClickHouseClientProtocol:
+    client = _make_clickhouse_client(settings)
+    if not isinstance(client, ClickHouseClientProtocol):
+        raise TypeError('clickhouse_driver.Client does not satisfy the ClickHouse client contract.')
```
bit-mis: p2: this `TypeError` branch is effectively unreachable. `@runtime_checkable` `Protocol` only validates attribute presence (`hasattr(obj, 'execute') and hasattr(obj, 'disconnect')`) — not signatures or return types — so `clickhouse_driver.Client` will always satisfy it. The check provides no real contract guarantee and risks misleading future readers into thinking it does. More broadly, `get_clickhouse_settings` / `make_clickhouse_client` here (and the new `database_exists` / `table_exists` wrappers in `create_binance_trades_table_origo.py`) just forward to their underscore-prefixed counterparts while `__all__` exports both versions side-by-side, doubling the public API for identical behavior and leaving new callers guessing which is canonical. Suggest either dropping the wrappers and importing the underscore helpers directly (matching the `create_binance_spot_dollar_klines_table_origo.py` pattern that the prior reviewers pointed to), or renaming the underscore helpers to drop the prefix and removing the duplicates.

bit-mis (changes requested):

bit-mis (approved):

PR #215, opened 2026-05-20
fix(origo): add manual backfill jobs

mikkokotila:
Closes #214

## Summary
- Add manual Origo backfill job for raw Binance spot trades only.
- Add minute-partitioned Origo depth20 backfill job for snapshots plus 1m projection.
- Convert depth20 minute selection from run config to Dagster partition keys starting at `2026-05-14T10:28:00+0000`.
- Add registration tests plus version/changelog.

## Validation
- `/tmp/tdw-control-plane-venv/bin/python -m pytest tests/origo_source_native -q --maxfail=1`
- `uvx ruff==0.15.11 check tools tests/tools`
- `git diff --check`
- `/tmp/tdw-control-plane-venv/bin/python tools/fail_loud_gate.py --base-budget .github/fail_loud_budget.json`
- `/tmp/tdw-control-plane-venv/bin/python tools/typing_gate.py --pyright-json /tmp/manual-backfill-jobs-pyright-2.json --base-budget .github/typing_budget.json`
- `/tmp/tdw-control-plane-venv/bin/python tools/version_gate.py ...`
- `/tmp/tdw-control-plane-venv/bin/python tools/slice_gate.py ...`
- `/tmp/tdw-control-plane-venv/bin/python tools/cc_gate.py ...`

### DIFF
--- a/tdw_control_plane/definitions.py
+++ b/tdw_control_plane/definitions.py
@@ -128,6 +128,7 @@ from .assets.refresh_aligned_1m_exchange_from_binance_futures_origo import (
     refresh_aligned_1m_exchange_from_binance_futures_origo,
 )
 from .assets.sync_binance_spot_depth20_snapshots_to_origo import (
+    depth20_minute_partitions,
     sync_binance_spot_depth20_snapshots_to_origo,
 )
 
@@ -266,6 +267,11 @@ backfill_binance_spot_trades_origo_job = define_asset_job(
     selection=['insert_daily_binance_spot_trades_to_origo'],
 )
 
+_BINANCE_SPOT_DEPTH20_DATA_SOURCE_SELECTION = [
+    'sync_binance_spot_depth20_snapshots_to_origo',
+    'refresh_binance_spot_depth20_1m_origo',
+]
+
 refresh_binance_futures_data_source_job = define_asset_job(
     name="refresh_binance_futures_data_source_job",
     selection=[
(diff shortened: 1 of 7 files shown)


zero-bang (approved):
LGTM — clean, well-scoped addition. Both jobs follow the established backfill pattern. Tests verify exact node membership and partition semantics. No missing consumers or configuration gaps.

tdw_control_plane/definitions.py
```diff
@@ -277,6 +282,14 @@ class _AssetEventLike(Protocol):
     ],
 )
 
+backfill_binance_spot_depth20_data_source_job = define_asset_job(
+    name='backfill_binance_spot_depth20_data_source_job',
+    selection=[
+        'sync_binance_spot_depth20_snapshots_to_origo',
+        'refresh_binance_spot_depth20_1m_origo',
+    ],
```
bit-mis: P2: This `selection` list is byte-for-byte identical to `refresh_binance_spot_depth20_data_source_job` defined ten lines above (same two asset keys, same order). The two jobs are not subsets of each other and are not differentiated by their selection at all — only by name and by the fact that one has a schedule. That makes the duplication a maintenance hazard: when the depth20 data-source pipeline gains or loses an asset, both lists must be updated in lockstep, and there is no compile-time or test-time check that they stay aligned. Consider extracting the list to a module-level constant (e.g. `_BINANCE_SPOT_DEPTH20_DATA_SOURCE_SELECTION = ['sync_binance_spot_depth20_snapshots_to_origo', 'refresh_binance_spot_depth20_1m_origo']`) and passing it to both `define_asset_job` calls so the refresh job and the manual backfill job cannot silently drift apart.

bit-mis (changes requested):

zero-bang (approved):

bit-mis (approved):

zero-bang (approved):

PR #217, opened 2026-05-20
fix(origo): reconcile depth20 partition state

mikkokotila:
## What does this PR change?

Adds a manual Dagster job that reconciles depth20 partition materialization state from existing Origo ClickHouse rows. It records runless Dagster materializations for existing snapshot and 1m projection minutes; it does not insert, delete, or update ClickHouse data.

Root cause: PR #215 made depth20 assets partitioned, but historical ClickHouse rows had no matching Dagster partition events, so Dagit showed all existing minutes as missing.

Closes #216

## Validation

- `python -m pytest tests/origo_source_native -q --maxfail=1`
- `uvx ruff==0.15.11 check tools tests/tools`
- `python tools/fail_loud_gate.py --base-budget .github/fail_loud_budget.json`
- `python tools/typing_gate.py --pyright-json /tmp/depth20-reconcile-pyright.json --base-budget .github/typing_budget.json`
- `python tools/version_gate.py --pr-title "fix(origo): reconcile depth20 partition state" --base-pyproject /tmp/depth20-base-pyproject.toml --head-pyproject pyproject.toml --base-changelog /tmp/depth20-base-changelog.md --head-changelog CHANGELOG.md`

## Checklist

- [x] I ran the project's test suite locally without errors (where applicable)
- [ ] I added and/or updated docstrings as per [Writing Docstrings](https://github.com/Vaquum/Limen/blob/main/docs/Developer/Writing-Docstrings.md) (for any changed public functions/classes)

### DIFF
--- a/tdw_control_plane/assets/reconcile_binance_spot_depth20_partition_state_origo.py
+++ b/tdw_control_plane/assets/reconcile_binance_spot_depth20_partition_state_origo.py
@@ -51,7 +51,7 @@ def _existing_snapshot_minutes(client: ClickHouseClient, database: str) -> list[
         client.execute(
             f"""
             SELECT DISTINCT toStartOfMinute(datetime) AS minute
-            FROM {database}.{SNAPSHOTS_TABLE_NAME} FINAL
+            FROM {database}.{SNAPSHOTS_TABLE_NAME}
             ORDER BY minute
             """
         )
@@ -63,7 +63,7 @@ def _existing_projection_minutes(client: ClickHouseClient, database: str) -> lis
         client.execute(
             f"""
             SELECT DISTINCT datetime AS minute
-            FROM {database}.{DEPTH20_1M_TABLE_NAME} FINAL
+            FROM {database}.{DEPTH20_1M_TABLE_NAME}
             ORDER BY minute
             """
         )
--- a/tdw_control_plane/assets/reconcile_binance_spot_depth20_partition_state_origo.py
+++ b/tdw_control_plane/assets/reconcile_binance_spot_depth20_partition_state_origo.py
@@ -1,0 +1,137 @@
+from collections.abc import Sequence
+from datetime import datetime, timezone
+from typing import cast
+
+from dagster import AssetExecutionContext, AssetKey, AssetMaterialization, asset
+
+from .create_binance_spot_depth20_1m_table_origo import DEPTH20_1M_TABLE_NAME
+from .create_binance_spot_depth20_snapshots_table_origo import (
+    ClickHouseClient,
+    SNAPSHOTS_TABLE_NAME,
+    get_clickhouse_settings,
+    make_clickhouse_client,
+)
+from .refresh_binance_spot_depth20_1m_origo import refresh_binance_spot_depth20_1m_origo
+from .sync_binance_spot_depth20_snapshots_to_origo import (
+    DEPTH20_PARTITION_KEY_FORMAT,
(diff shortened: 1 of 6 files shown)


zero-bang (approved):
LGTM — clean, well-scoped reconciliation job.

**What I verified:**
- All imports resolve to existing exports; no circular dependencies.
- `_partition_key` correctly handles naive ClickHouse datetimes by assuming UTC, producing keys matching `DEPTH20_PARTITION_KEY_FORMAT` (`%Y-%m-%dT%H:%M:%S%z`).
- Idempotency is correct: `_report_existing_partitions` checks `get_materialized_partitions` before reporting, so repeated runs are safe no-ops.
- The reconcile job is unpartitioned and has no schedule — manual-only, as intended.
- All four registrations in `definitions.py` are present (import, job def, assets list, jobs list).
- The reconcile asset is correctly excluded from `_BINANCE_SPOT_DEPTH20_DATA_SOURCE_SELECTION`, so it won't fire on the per-minute schedule.
- SQL uses `FINAL` for merge-tree correctness; `toStartOfMinute` for sub-minute snapshot precision vs. direct `datetime` for the already-minute-level 1m table.
- `client.disconnect()` in `finally` block ensures cleanup.
- Test covers the happy path end-to-end: table setup, row insertion, reconcile materialization, and partition state assertion.
- No SQL injection risk — table names and database come from constants/env config, not user input.
- CHANGELOG and version bump are correct.

**Forward-looking note:** if ClickHouse ever accumulates enough depth20 history that the `SELECT DISTINCT ... FINAL` scan becomes slow, consider batching by date range. Not a concern at current scale.

tdw_control_plane/assets/reconcile_binance_spot_depth20_partition_state_origo.py
```diff
@@ -0,0 +1,137 @@
...
+        minute = minute.replace(tzinfo=timezone.utc)
+    return minute.astimezone(timezone.utc).strftime(DEPTH20_PARTITION_KEY_FORMAT)
+
+
+def _existing_snapshot_minutes(client: ClickHouseClient, database: str) -> list[datetime]:
+    return _datetime_rows(
+        client.execute(
+            f"""
+            SELECT DISTINCT toStartOfMinute(datetime) AS minute
+            FROM {database}.{SNAPSHOTS_TABLE_NAME} FINAL
```
bit-mis: [p2] `FINAL` here (and at line 66 in `_existing_projection_minutes`) is unnecessary and adds substantial query cost. The result set is unchanged: `DISTINCT toStartOfMinute(datetime)` already collapses per-`source_timestamp_ms` duplicates into one minute key, and on the 1m table `DISTINCT datetime` collapses the same way (`datetime` is the `ORDER BY` key). Unlike the existing `FINAL` uses in `refresh_binance_spot_depth20_1m_origo.py` — which need it for accurate `count()` over a minute window and for latest-version selection during aggregation — reconciliation only needs minute-set membership.

On the snapshots table, which receives many writes per minute and (per the PR description) has accumulated historical rows since #215, `FINAL` forces ClickHouse to merge all `ReplacingMergeTree` parts at query time on top of a scan that already touches every row. For a one-shot reconcile potentially covering months of partitions, this is a meaningful regression with no functional benefit. Suggest dropping `FINAL` on both reconcile queries.

bit-mis (changes requested):

bit-mis (approved):

PR #219, opened 2026-05-21
fix(origo): correct dollar kline sizes

mikkokotila:
Closes #218

## Summary
- Set the Origo dollar kline base size to 1M.
- Replace the HF dollar kline publisher set with 1M, 15M, 30M, 60M, 120M, and 240M.
- Keep HF repo/file slugs uppercase `M`, matching the dataset labels.

## Explicit non-change
- No production `DROP`, `DELETE`, or table replacement operation is added.
- No raw trades, depth20, imbalance bars, schedules, or unrelated assets are changed.
- Replacing the already-built 100k-derived production table remains an explicit operator action after review/deploy.

## Validation
- `pytest tests/origo_source_native/test_origo_binance_spot_dollar_klines_template.py tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py -q --maxfail=1` passed: 12 passed.
- `pytest tests/origo_source_native -q --maxfail=1` passed: 80 passed.
- `git diff --check` passed.
- `uvx ruff==0.15.11 check tools tests/tools` passed.


zero-bang (approved):
LGTM — clean, consistent rename and base-size change.

- Verified no stale references to old asset names (100k, 2M, 4M, 8M, 16M, 32M) remain anywhere on the branch.
- `base_bar_count` math checks out: all new sizes are integer multiples of the 1M base.
- `DOLLAR_KLINE_SIZE` is a single source of truth imported by the HF utility — no hardcoded copies.
- New contract test locks in the 1M base size.
- Dataset card template updated correctly.

Forward-looking note: the next operational risk is running the new publishers against a production table that still holds 100k-based bars — but the PR description already calls this out as an explicit operator action, which is the right scoping.

zero-bang (approved):

tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py
```diff
@@ -214,7 +214,7 @@ def _origo_dollar_klines_dataframe(
     start_date_limit: str,
     end_date_limit: str,
 ) -> pl.DataFrame:
-    base_bar_count = int(dollar_size / 100_000.0)
+    base_bar_count = int(dollar_size / 1_000_000.0)
```
bit-mis: p2 — `_origo_dollar_klines_dataframe` hardcodes the base divisor as `1_000_000.0`, duplicating the production `DOLLAR_KLINE_SIZE` rather than importing it. This PR already changed this magic number once (from `100_000.0` to `1_000_000.0`) in lockstep with `refresh_binance_spot_dollar_klines_origo.DOLLAR_KLINE_SIZE`, which demonstrates the coupling is silent — nothing fails if the two drift. Note the inconsistency with the new `test_binance_spot_dollar_klines_base_size_contract` in `test_origo_binance_spot_dollar_klines_template.py`, which correctly references `refresh_binance_spot_dollar_klines_origo_module.DOLLAR_KLINE_SIZE` dynamically. Suggest importing `DOLLAR_KLINE_SIZE` (or reusing `_base_bar_count_for_dollar_size`) here so the test stays correct if the production base size is ever changed again; otherwise a future tweak to `DOLLAR_KLINE_SIZE` would have this helper silently compute the wrong `base_bar_count` while the surrounding assertions still pass.

bit-mis (changes requested):

zero-bang (approved):

bit-mis (approved):

PR #221, opened 2026-05-22
fix(origo): require raw spot trades for dollar imbalance refresh

mikkokotila:
## What does this PR change?

Closes #220.

Requires the dollar imbalance kline refresh to prove the raw Binance spot trades partition is materialized, present, and ledger-consistent before replacing existing derived rows.

This matches the fixed dollar kline standard: raw spot trades are a hard dependency, and missing raw input fails before stale derived rows can be deleted.

## Validation

- Focused dollar imbalance test file: `6 passed`
- Full `tests/origo_source_native`: `81 passed`
- Ruff: pass
- Version gate: pass
- Fail-loud gate: pass
- Typing gate: pass
- CC gate: pass
- Slice gate: pass locally with a single closing reference

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v1.14.2 on May 22, 2026
+- Require raw spot trades before replacing Binance spot dollar imbalance kline partitions.
+
 # v1.14.1 on May 21, 2026
 - Fix the Binance spot dollar kline base size to 1M and publish 1M, 15M, 30M, 60M, 120M, and 240M dollar snapshots.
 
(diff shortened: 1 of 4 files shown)


zero-bang (approved):
LGTM. The four validation functions (`_count_raw_partition_rows`, `_raw_ledger_inserted_row_count`, `_raw_partition_was_materialized`, `_ensure_raw_partition_ready`) correctly mirror the established pattern from `refresh_binance_spot_dollar_klines_origo.py`. The `_ensure_raw_partition_ready` call is correctly placed before the delete-and-replace logic, preserving existing derived rows when raw input is missing. The test follows the same structure as the dollar klines analog and verifies the core safety invariant: existing imbalance kline rows survive a failed refresh.

PR #227, opened 2026-06-04
fix(huggingface): publish time klines from Origo projection

mikkokotila:
## Summary
- Route Hugging Face spot time-kline exports through `origo.binance_spot_klines`.
- Move the six time-kline publish sensors to `refresh_binance_spot_klines_origo`.
- Keep HF file names, repos, metadata shape, and exported columns unchanged.

Closes #226

## Validation
- `ruff check tools tests/tools`
- `ruff check --fix tdw_control_plane/utils/publish_binance_spot_kline_snapshot_to_huggingface.py`
- `/tmp/tdw-control-plane-venv/bin/python -m compileall tdw_control_plane tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py tests/origo_source_native/test_publish_binance_spot_additional_klines_to_huggingface.py tests/origo_source_native/test_publish_binance_spot_1h_klines_to_huggingface.py tests/origo_source_native/test_publish_binance_spot_4h_klines_to_huggingface.py`
- `pyright --pythonpath /tmp/tdw-control-plane-venv/bin/python tdw_control_plane/utils/publish_binance_spot_kline_snapshot_to_huggingface.py`
- fake-client projection helper check against `origo.binance_spot_klines`
- `/tmp/tdw-control-plane-venv/bin/python -m pytest --collect-only tests/origo_source_native/test_publish_binance_spot_klines_to_huggingface.py tests/origo_source_native/test_publish_binance_spot_additional_klines_to_huggingface.py tests/origo_source_native/test_publish_binance_spot_1h_klines_to_huggingface.py tests/origo_source_native/test_publish_binance_spot_4h_klines_to_huggingface.py -q`

## Local Blocker
- Full Origo pytest execution is blocked locally because Docker daemon is not running: `connect: no such file or directory` for `/Users/mikkokotila/.docker/run/docker.sock`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v1.15.2 on June 4, 2026
+- Route Hugging Face Binance spot time-kline exports through the Origo 1m kline projection.
+
 # v1.15.1 on May 30, 2026
 - Include latest Binance spot table creation in the scheduled latest data-source job.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "quickstart_etl"
-version = "1.15.1"
+version = "1.15.2"
 description = "Trades Warehouse Control Plane"
 readme = "README.md"
 requires-python = ">=3.11"
--- a/tdw_control_plane/definitions.py
+++ b/tdw_control_plane/definitions.py
@@ -766,7 +766,7 @@ def _publish_binance_spot_dollar_klines_to_hf_run_request(
 
 
 @asset_sensor(
-    asset_key=AssetKey("insert_daily_binance_spot_trades_to_origo"),
+    asset_key=AssetKey("refresh_binance_spot_klines_origo"),
     job=publish_binance_spot_klines_to_huggingface_job,
 )
 def publish_binance_spot_klines_to_huggingface_sensor(
@@ -780,7 +780,7 @@ def publish_binance_spot_klines_to_huggingface_sensor(
 
 
 @asset_sensor(
-    asset_key=AssetKey("insert_daily_binance_spot_trades_to_origo"),
+    asset_key=AssetKey("refresh_binance_spot_klines_origo"),
     job=publish_binance_spot_15m_klines_to_huggingface_job,
 )
 def publish_binance_spot_15m_klines_to_huggingface_sensor(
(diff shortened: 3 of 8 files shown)


zero-bang (approved):
LGTM — this is a clean, faithful port that routes the six time-kline HF exports through `origo.binance_spot_klines` and mirrors the already-merged dollar-kline sibling.

**Verified the rollup math is exact (not just plausible):**
- **Bucket alignment**: re-bucketing the 1m projection by an N×60s size gives `3600·floor(floor(t/60)/60) = 3600·floor(t/3600)` — identical to bucketing raw trades directly, so no boundary drift for any multiple of 60.
- **mean**: `sum(n·mean)/sum(n)` reconstructs `avg(price)` exactly, since the projection stores `mean = avg(price)` per minute.
- **std**: the pooled population-variance formula `E[X²]−E[X]²` exactly reconstructs `stddevPopStable(price)` because the projection stores population std (`refresh_binance_spot_klines_origo` uses `stddevPopStable`). The `greatest(…, 0)` guard covers float cancellation.
- **OHLC** via `argMin/argMax(…, source_datetime)` correctly picks the first/last minute's open/close, and volume/trade-count/liquidity/maker_ratio are exact additive or count-weighted reconstructions. The 1m export (kline_size=60) re-buckets as an identity. Projection `mean`/`std` are stored unrounded, so no precision is lost feeding the aggregation.

**Wiring checks out**: `refresh_binance_spot_klines_origo` is a scheduled asset (in `refresh_binance_spot_data_source_job`) that depends on `insert_daily_binance_spot_trades_to_origo`, so moving the sensors onto it correctly sequences the export *after* its actual source is refreshed. Dollar-kline sensors correctly stay on the trades asset. All six cadences (1m/15m/30m/1h/2h/4h), the dataset card, CHANGELOG, and version are updated consistently.

**Non-blocking observations** (no change required):
- The std rollup subtracts two large near-equal quantities (E[X²] vs E[X]²) at BTC price magnitudes (~1e10). Float64 resolves this comfortably for realistic intra-window variance, and the `greatest(…,0)` guard matches the sibling — just flagging it as the one spot where projection-sourced std can diverge in the last digits from a direct trades computation.
- The tests fully monkeypatch `_get_binance_spot_klines_from_1m_projection`, so the production SQL string and the `.cast(Datetime("ms","UTC"))` path aren't exercised by unit tests. This matches the repo's existing pattern, but it means the actual query is only validated by the (docker-blocked) integration run.

**Forward-looking**: the next predictable failure mode is a schema/semantics drift between the 1m projection and this rollup — e.g. if `refresh_binance_spot_klines_origo` ever switches `std` to `stddevSamp` or rounds `mean` in-table, the pooled-variance reconstruction here would silently skew without any test catching it. A small integration assertion comparing a projection rollup against a direct trades-based kline for one day would close that gap for both this and the dollar-kline helper.

zero-bang (approved):

PR #211, opened 2026-05-20
feat(huggingface): publish Binance spot dollar klines

mikkokotila:
Closes #210

## Validation
- `python -m compileall tdw_control_plane/utils/publish_binance_spot_dollar_kline_snapshot_to_huggingface.py tdw_control_plane/assets/publish_binance_spot_100k_dollar_klines_to_huggingface.py tdw_control_plane/assets/publish_binance_spot_2m_dollar_klines_to_huggingface.py tdw_control_plane/assets/publish_binance_spot_4m_dollar_klines_to_huggingface.py tdw_control_plane/assets/publish_binance_spot_8m_dollar_klines_to_huggingface.py tdw_control_plane/assets/publish_binance_spot_16m_dollar_klines_to_huggingface.py tdw_control_plane/assets/publish_binance_spot_32m_dollar_klines_to_huggingface.py tdw_control_plane/definitions.py tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py`
- `pytest tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py -q`
- `pytest tests/origo_source_native -q --maxfail=1`
- `pyright --outputjson` + `tools/typing_gate.py --base-budget .github/typing_budget.json --pyright-json /tmp/pyright-issue-210.json`
- `tools/fail_loud_gate.py --base-budget .github/fail_loud_budget.json`
- `ruff check tools tests/tools`
- `tools/slice_gate.py ...`
- `tools/version_gate.py ...`
- `tools/cc_gate.py ...`

PR #232, opened 2026-06-04
feat(origo): mirror Binance spot klines to a local parquet mount

mikkokotila:
Closes #229

## Summary
A stateless 10-minute job mirrors the 12 HuggingFace Binance spot kline series to **monthly** Parquet files on a local mount (`LOCAL_PARQUET_DIR`, default `/opt/parquet`), built by a factory over a 12-row spec. Additive to the existing daily HF publish jobs.

- **`query/binance_spot_kline_rollups.py`** — monthly `time_month` / `dollar_month` rollups. Time rolls the 1m base `UNION ALL` the rolling-latest 1m (one query, `base_cut` as a scalar CTE so the daily refresh can't land a partition mid-read). Dollar takes finalized days from the day-scoped base plus the still-open day(s) recomputed from the rolling raw trades with the same within-day running-$-sum the base build uses. Reuses the #227/#211 aggregation contracts.
- **`assets/publish_binance_spot_klines_to_mount.py`** — spec + factory (12 assets), a monthly atomic writer with a **monotonic guard** (a slow, out-of-order tick can never replace fresher data with staler — atomic ≠ monotonic), and `tick` / `backfill` modes.
- **`definitions.py`** — the 12 assets + a `*/10` tick job and a one-shot `backfill` job (both `in_process_executor`), registered with the schedule.
- **compose** — additive external `tdw-parquet` volume (prod) and a dev bind mount; existing `clickhouse-data`/`dagster-instance` mounts untouched.

Layout: `/opt/parquet/{family}/{interval}/{YYYY}/{MM}.parquet` (aligns with the 1m base's `toYYYYMM` partitions).

## Validation (local, against an ephemeral ClickHouse via the conftest harness)
- `pytest tests/origo_source_native -q` → **99 passed** (incl. 7 new tests: factory, schedule, time/dollar parity, monotonic writer, atomic round-trip, dollar open-day-from-raw).
- pyright (strict): **error-count delta 0** vs `origin/main`; the two new modules have 0 errors.
- The dollar open-day path is exercised end-to-end (base = day 1, rolling raw = day 2) and matches the canonical base build.

## Deploy (no manual host step)
The `tdw-parquet` volume is Compose-managed (name-pinned, **not** `external`), so `docker compose up` creates `tdw-control-plane_parquet` automatically on deploy — the repo fully defines it, no out-of-band `docker volume create` on the host. It holds reproducible mirror output (rebuildable via the backfill job), so it doesn't need the external protection the clickhouse-data / dagster-instance volumes have. Historical months are filled once by launching the repo-defined `backfill_binance_spot_klines_to_mount_job` in Dagit; the `*/10` schedule (RUNNING) keeps the open month fresh after that.

## Out-of-scope finding worth a follow-up
While building the dollar parity test I found that `_get_binance_spot_dollar_klines` (the HF dollar export, #211) returns **1970** `start_datetime`/`end_datetime` under polars ≥ 1.40 — `pl.from_arrow` doesn't rescale ClickHouse second-precision timestamps, and that export `cast`s straight to `Datetime("ms")` without the `*1000` compensation the time export uses. The existing dollar test masks it by monkeypatching the function out. `dollar_month` here avoids it (its union promotes to `DateTime64(3)`), but the live HF dollar files may carry 1970 timestamps. Flagged for a separate fix — out of this slice's scope.

### DIFF
--- a/tdw_control_plane/assets/publish_binance_spot_klines_to_mount.py
+++ b/tdw_control_plane/assets/publish_binance_spot_klines_to_mount.py
@@ -8,6 +8,7 @@ file with staler data (atomic != monotonic). A separate ``backfill`` mode
 rebuilds every month from 2020-01 for the initial fill or after a gap.
 """
 
+import fcntl
 import os
 import time
 import uuid
@@ -19,7 +20,11 @@ from typing import Literal
 import polars as pl
 from dagster import AssetExecutionContext, AssetsDefinition, Config, asset
 
-from tdw_control_plane.query.binance_spot_kline_rollups import dollar_month, time_month
+from tdw_control_plane.query.binance_spot_kline_rollups import (
+    dollar_month,
+    dollar_open_day_gap_days,
+    time_month,
+)
 
 DEFAULT_MOUNT_DIR = "/opt/parquet"
 EXPORT_START_YEAR = 2020
@@ -52,7 +57,9 @@ SPECS: tuple[MountKlineSpec, ...] = (
 
 
 class MountExportConfig(Config):
-    mode: str = "tick"  # "tick" -> open month(s); "backfill" -> every month from 2020-01
+    # "tick" -> open month(s); "backfill" -> every month from 2020-01.
+    # Literal so Dagster/pydantic validates the value at launch time.
+    mode: Literal["tick", "backfill"] = "tick"
     dry_run: bool = False
 
 
@@ -124,8 +131,10 @@ def _clear_orphan_temp_files(directory: Path, month: int) -> None:
 
 
 def write_month_atomic(df: pl.DataFrame, spec: MountKlineSpec, year: int, month: int) -> str:
-    """Write a month file atomically (temp in same dir -> os.replace), but skip
-    the replace when the existing file already holds data at least as fresh.
+    """Write a month file atomically (temp in same dir -> os.replace), skipping the
+    replace when the existing file already holds data at least as fresh. A per-month
+    file lock serializes the freshness check and the replace, so overlapping ticks
+    cannot race and regress freshness (staler-over-fresher).
     Returns ``written``, ``skipped_empty``, or ``skipped_not_newer``."""
     if df.height == 0:
         return "skipped_empty"
@@ -137,17 +146,23 @@ def write_month_atomic(df: pl.DataFrame, spec: MountKlineSpec, year: int, month:
 
     target = month_path(spec.sub_path, year, month)
     target.parent.mkdir(parents=True, exist_ok=True)
-
-    if target.exists():
-        existing_max = _existing_max_timestamp(target, column)
-        if existing_max is not None and existing_max >= new_max:
-            return "skipped_not_newer"
-
     _clear_orphan_temp_files(target.parent, month)
     tmp = target.parent / f".{month:02d}.parquet.partial-{os.getpid()}-{uuid.uuid4().hex}"
     df.write_parquet(tmp, compression="zstd")
-    os.replace(tmp, target)
-    return "written"
+
+    # Hold a per-month lock across the check-and-replace so two overlapping ticks
+    # cannot both read the same stale max and race the replace. The temp write above
+    # stays outside the lock.
+    lock_path = target.parent / f".{month:02d}.parquet.lock"
+    with open(lock_path, "w", encoding="utf-8") as lock_file:
+        fcntl.flock(lock_file, fcntl.LOCK_EX)
+        if target.exists():
+            existing_max = _existing_max_timestamp(target, column)
+            if existing_max is not None and existing_max >= new_max:
+                tmp.unlink()
+                return "skipped_not_newer"
+        os.replace(tmp, target)
+        return "written"
 
 
 def run_mount_export(
@@ -155,6 +170,17 @@ def run_mount_export(
 ) -> dict[str, object]:
     now = datetime.now(UTC)
     months = months_for_run(config.mode, now)
+
+    if spec.family == "dollar" and not config.dry_run:
+        gap_days = dollar_open_day_gap_days()
+        if gap_days > 0:
+            context.log.warning(
+                f"{spec.name}: the dollar base is {gap_days} day(s) behind what the 2-day "
+                "rolling raw table can cover; those days are absent from both sources, so the "
+                "affected month file will have a hole until refresh_binance_spot_dollar_klines_origo "
+                "catches up or a backfill is run."
+            )
+
     results: dict[str, str] = {}
     for year, month in months:
         key = f"{year:04d}-{month:02d}"
--- a/tdw_control_plane/assets/publish_binance_spot_klines_to_mount.py
+++ b/tdw_control_plane/assets/publish_binance_spot_klines_to_mount.py
@@ -1,0 +1,194 @@
+"""Mirror the 12 HF Binance spot kline series to monthly Parquet files on a
+local mount, refreshed every 10 minutes.
+
+Stateless: each tick rebuilds the open month(s) from the Origo projections and
+replaces the month file atomically. The filesystem is the only state. A
+monotonic guard ensures a slow, out-of-order tick can never replace a fresher
+file with staler data (atomic != monotonic). A separate ``backfill`` mode
+rebuilds every month from 2020-01 for the initial fill or after a gap.
+"""
+
+import os
+import time
+import uuid
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from pathlib import Path
+from typing import Literal
+
+import polars as pl
+from dagster import AssetExecutionContext, AssetsDefinition, Config, asset
+
+from tdw_control_plane.query.binance_spot_kline_rollups import dollar_month, time_month
+
+DEFAULT_MOUNT_DIR = "/opt/parquet"
+EXPORT_START_YEAR = 2020
+EXPORT_START_MONTH = 1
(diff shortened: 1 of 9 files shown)


tests/origo_source_native/test_publish_binance_spot_klines_to_mount.py
```diff
@@ -0,0 +1,277 @@
...
+def test_mount_export_schedule_runs_every_ten_minutes(origo_definitions_module: Any) -> None:
+    schedule = origo_definitions_module.publish_binance_spot_klines_to_mount_schedule
+    assert schedule.cron_schedule == "*/10 * * * *"
+    assert schedule.execution_timezone == "UTC"
+
+    assert months_for_run("tick", datetime(2024, 3, 15)) == [(2024, 2), (2024, 3)]
+    assert months_for_run("backfill", datetime(2020, 3, 1)) == [(2020, 1), (2020, 2), (2020, 3)]
+
+
+def test_time_month_rollup_matches_hf_projection(origo_assets: dict[str, Any]) -> None:
```
zero-bang, line 109: The time path's one novel behavior over the HF projection — the `base <= base_cut` UNION `latest > base_cut` splice — is never exercised. `_materialize_time_base` (L57) refreshes only `binance_spot_klines`; `binance_spot_klines_latest` is created but never populated, so `base_cut = max(base)` makes the `latest > base_cut` branch contribute zero rows and this test reduces to the plain base aggregation (identical to the HF projection it compares against). The symmetric dollar behavior got a real test (`test_dollar_open_day_from_raw_matches_base_refresh`); the time splice deserves the same. Suggest: materialize `refresh_binance_spot_latest_1m` (or insert directly into `binance_spot_klines_latest`) for minutes strictly after the base's max datetime, then assert `time_month` returns those rolling minutes and that they are spliced at the `base_cut` boundary with no gap/overlap.

zero-bang (commented):
Reviewed the rollup queries, the atomic/monotonic writer, the factory + job wiring, and the compose/volume changes. This is a careful, well-documented slice: the dollar open-day-from-raw path has real end-to-end parity coverage, the atomic writer is TOCTOU-narrowed with a monotonic guard, and the union/scalar-CTE splices match the #227/#211 contracts. The out-of-scope #211 1970-timestamp flag is a good catch and correctly avoided here (the DateTime64(3) promotion). pyright-clean and additive.

Submitting as COMMENT (not REQUEST_CHANGES) — I did not find a correctness bug. Two findings worth addressing before merge plus one engineering note are inline. The main one is a test gap: the time path's distinguishing new behavior (the base_cut UNION with the rolling-latest table) is never exercised, while the symmetric dollar path got a dedicated test. The second is a degraded-mode data-hole risk tied to the 2-day raw TTL.

Forward-looking failure mode: the next predictable break is the dollar base refresh (refresh_binance_spot_dollar_klines_origo) stalling for >2 days — see the inline comment on the raw WHERE clause. A finalized prev-month is also fully re-queried every 10 min only to be skipped at write time (the guard saves the write, not the ClickHouse scan); acceptable, but worth knowing the steady-state cost is 24 month-aggregations per tick.

tdw_control_plane/query/binance_spot_kline_rollups.py
```diff
@@ -0,0 +1,401 @@
...
+                                    PARTITION BY toDate(datetime)
+                                    ORDER BY datetime, trade_id
+                                    ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
+                                ) - quote_quantity,
+                                0.0
+                            ) AS running_quote_before
+                        FROM {database}.{raw_latest_table}
+                        WHERE datetime >= toDateTime64({{start_dt:String}}, 3)
+                          AND datetime < toDateTime64({{end_dt:String}}, 3)
+                          AND toDate(datetime) > base_day
```
zero-bang, line 371: Open-day recompute reads `binance_spot_trades_latest`, which has a 2-day TTL (`create_binance_spot_latest_tables_origo.py:68`, `INTERVAL 2 DAY DELETE`). In steady state (dollar base at most ~1 day behind) the open day is always within the window, so this is fine. But if `refresh_binance_spot_dollar_klines_origo` stalls for >2 days, the days in `(base_day, now-2d]` are absent from BOTH sources — not yet finalized in the base (`toDate <= base_day` excludes them) and already TTL'd from raw — so the month file silently develops a hole rather than failing loud. Once the stall exceeds a month boundary the affected month also drops out of `months_for_run('tick')`, so it won't self-heal until a manual `backfill`. Worth at least a guard/log when `min(open-day) > base_day + 1` (i.e. raw cannot cover the gap), so the degraded mode is observable instead of silent.

tdw_control_plane/assets/publish_binance_spot_klines_to_mount.py
```diff
@@ -0,0 +1,194 @@
...
+    for year, month in months:
+        key = f"{year:04d}-{month:02d}"
+        if config.dry_run:
+            results[key] = "dry_run"
+            continue
+        df = _build_month_df(spec, year, month)
+        outcome = write_month_atomic(df, spec, year, month)
+        results[key] = outcome
+        context.log.info(f"{spec.name} {key}: {outcome} ({df.height} rows)")
+
```
zero-bang, line 194: Engineering note (non-blocking): the monotonic guard is best-effort under truly concurrent ticks. The sequence `target.exists()` -> `_existing_max_timestamp` -> `os.replace` is not atomic, so two overlapping runs can both read the same stale `existing_max`, both pass the guard, and the one that calls `os.replace` last wins — re-introducing the exact staler-over-fresher regression the guard defends against. With a 10-min cron and `in_process_executor` this window is small and likely never hit, but Dagster doesn't prevent overlapping schedule ticks by default. If you want the guard to be airtight, a per-series file lock around the read-check-replace (or a run-concurrency limit of 1 on the job) would close it. Fine to leave as-is given the cadence — just flagging that the docstring's monotonicity claim holds for out-of-order *completion* of non-overlapping runs, not for genuinely concurrent ones.

tdw_control_plane/assets/publish_binance_spot_klines_to_mount.py
```diff
@@ -0,0 +1,194 @@
...
+    if target.exists():
+        existing_max = _existing_max_timestamp(target, column)
+        if existing_max is not None and existing_max >= new_max:
+            return "skipped_not_newer"
+
+    _clear_orphan_temp_files(target.parent, month)
+    tmp = target.parent / f".{month:02d}.parquet.partial-{os.getpid()}-{uuid.uuid4().hex}"
+    df.write_parquet(tmp, compression="zstd")
+    os.replace(tmp, target)
+    return "written"
```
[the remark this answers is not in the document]
mikkokotila: Fixed in d464627 — `write_month_atomic` now holds a per-month `fcntl.flock` across the freshness check + `os.replace`, so a slow/concurrent tick re-reads the current max under the lock right before replacing and cannot overwrite fresher data.

tdw_control_plane/assets/publish_binance_spot_klines_to_mount.py
```diff
@@ -0,0 +1,194 @@
...
+    MountKlineSpec("dollar_30M", "dollar", 30, "dollar/30M"),
+    MountKlineSpec("dollar_60M", "dollar", 60, "dollar/60M"),
+    MountKlineSpec("dollar_120M", "dollar", 120, "dollar/120M"),
+    MountKlineSpec("dollar_240M", "dollar", 240, "dollar/240M"),
+)
+
+
+class MountExportConfig(Config):
+    mode: str = "tick"  # "tick" -> open month(s); "backfill" -> every month from 2020-01
+    dry_run: bool = False
```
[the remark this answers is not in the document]
mikkokotila: Done in d464627 — `mode` is now `Literal["tick", "backfill"]`, so pydantic/Dagster validates it at launch time instead of failing later in `months_for_run`.

tests/origo_source_native/test_publish_binance_spot_klines_to_mount.py
mikkokotila: Good catch — added `test_time_month_splices_base_and_rolling_latest` (d464627): it materializes day 1 + day 2 into the base, moves day 2 into `binance_spot_klines_latest`, deletes it from the base, then asserts `time_month` reconstructs the full month — so day 2 can only arrive via the `base_cut` UNION latest splice — with no duplicate interval at the seam.

tdw_control_plane/query/binance_spot_kline_rollups.py
mikkokotila: Addressed in d464627 — added `dollar_open_day_gap_days()`; `run_mount_export` now logs a warning when the dollar base has stalled past the 2-day rolling-raw window (i.e. raw cannot cover `(base_day, ...]`), so the degraded mode is observable instead of a silent hole. Guarded to not fire on an empty base (fresh deploy).

tdw_control_plane/assets/publish_binance_spot_klines_to_mount.py
mikkokotila: Closed it rather than leaving best-effort — per-month `fcntl.flock` around the check-and-replace in d464627 (the file-lock option you suggested). Docstring updated so the monotonicity claim now holds for genuinely concurrent ticks, not just out-of-order completion.

zero-bang (approved):

PR #235, opened 2026-06-04
fix(huggingface): preserve dollar kline export timestamps

### LINKED ISSUE #234: fix(huggingface): preserve dollar kline export timestamps
## Slice ID

S233

## Slice Title

fix(huggingface): preserve dollar kline export timestamps

## Routing

Parent issue:
Depends on:
- none
Unlocks:
- none

## Minimum Viable Capability

>
> Accepted:
>
> Rejected:
> - *Strict CQRS separation.*

- `python3 -m pytest "tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py::test_dollar_kline_export_preserves_clickhouse_timestamp_scale" -q` exits `0`
- `python3 -m pytest "tests/origo_source_native/test_publish_binance_spot_klines_to_mount.py::test_dollar_month_rollup_matches_hf_day_scoped" -q` exits `0`

## Current State

>
> Accepted:
>
> Rejected:

- `grep -c 'toDateTime64' tdw_control_plane/utils/publish_binance_spot_dollar_kline_snapshot_to_huggingface.py` returns `0`
- `grep -c 'test_dollar_kline_export_preserves_clickhouse_timestamp_scale' tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py` returns `0`
- `grep -c '^# v1.16.1 ' CHANGELOG.md` returns `0`

## Target State

> predicate.
>
> Accepted:

- `grep -c 'toDateTime64(min(source_start_datetime), 3)' tdw_control_plane/utils/publish_binance_spot_dollar_kline_snapshot_to_huggingface.py` returns `1`
- `python3 -m pytest "tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py::test_dollar_kline_export_preserves_clickhouse_timestamp_scale" -q` exits `0`
- `python3 -m pytest tests/origo_source_native -q` exits `0`

## Surfaces

- `tdw_control_plane/utils/publish_binance_spot_dollar_kline_snapshot_to_huggingface.py`
- `tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py`
- `tests/origo_source_native/test_publish_binance_spot_klines_to_mount.py`
- `CHANGELOG.md`
- `pyproject.toml`

Added:
- none

Removed:
- none

## Signature Diff

### tdw_control_plane/utils/publish_binance_spot_dollar_kline_snapshot_to_huggingface.py::_get_binance_spot_dollar_klines (timestamp projection)
Before:
```sql
min(source_start_datetime) AS start_datetime,
max(source_end_datetime) AS end_datetime,
```
After:
```sql
toDateTime64(min(source_start_datetime), 3) AS start_datetime,
toDateTime64(max(source_end_datetime), 3) AS end_datetime,
```

### pyproject.toml::project.version
Before:
```toml
version = "1.16.0"
```
After:
```toml
version = "1.16.1"
```

### CHANGELOG.md::v1.16.1
Before:
```markdown
# v1.16.0 on June 4, 2026
```
After:
```markdown
# v1.16.1 on June 4, 2026
- Fix Binance spot dollar-kline Hugging Face exports collapsing every timestamp to ~1970 under polars >=1.40 by emitting millisecond-precision DateTime64 so the Arrow round-trip preserves the real dates.
```

## Tests That Prove the MVC

|---|---|---|
| HF dollar export keeps ClickHouse second-precision timestamps (no ~1970 collapse) under polars >=1.40 | `tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py` | `test_dollar_kline_export_preserves_clickhouse_timestamp_scale` |
| dollar_month mirror and the HF dollar export agree on the full frame, datetimes included | `tests/origo_source_native/test_publish_binance_spot_klines_to_mount.py` | `test_dollar_month_rollup_matches_hf_day_scoped` |

## Regression Gate

>
> Accepted:
>   on non-zero exit.
>
> Rejected:

Workflow file: `.github/workflows/pr_checks_tests.yml`
Job name: `pr_checks_tests`
Specific failing command the job runs: `pytest tests/origo_source_native -q --maxfail=1`

## Out of Scope

- `tdw_control_plane/utils/publish_binance_spot_kline_snapshot_to_huggingface.py`
- `tdw_control_plane/utils/asset_insert_to_tdw.py`
- `tdw_control_plane/query/binance_spot_kline_rollups.py`
- `tdw_control_plane/query/get_binance_spot_klines.py`
- `tdw_control_plane/assets/refresh_binance_spot_dollar_klines_origo.py`
- `docker-compose.yml`
- `docker-compose.deploy.yml`

## Risks and Open Questions

Risks:
- none — the polars cast tail is unchanged; the fix only changes the wire precision the export requests, matching the `dollar_month` twin already validated in #232.

Open questions:
- The six live `vaquum/binance_btcusdt_*_dollar_klines` files keep their 1970 timestamps only until the next daily publish; those jobs fully replace the file via `delete_patterns`, so they self-heal on the next run (or a manual Dagit trigger) after deploy — no backfill required.

## Closeout

> bot verifies:

Merge SHA:
Merged PR number:
-

## Author Checks

mikkokotila:
Closes #234

## Summary
The Hugging Face dollar-kline export (`_get_binance_spot_dollar_klines`, all six `vaquum/binance_btcusdt_*_dollar_klines` series) selected `min(start_datetime)` / `max(end_datetime)` from second-precision `DateTime` columns. Under **polars ≥1.40**, `pl.from_arrow` no longer rescales second-precision ClickHouse timestamps, so those values arrive as **epoch seconds** and the export's `.cast(Datetime("ms"))` reinterprets them as **milliseconds** — collapsing every dollar bar to **~1970-01-19**. (The time export dodges this with `(col.cast(Int64) * 1000)`; the dollar export had no compensation.)

**Fix:** wrap both timestamp aggregates in `toDateTime64(_, 3)` so ClickHouse emits millisecond precision over the wire — the same handling the `dollar_month` mirror twin (added in #232) already uses. The polars cast tail is unchanged and is now value-preserving.

## Validation (local, ephemeral ClickHouse via the conftest harness, polars 1.41.2)
- `pytest tests/origo_source_native` → **101 passed** (incl. the new regression test).
- **A/B proof the guard catches it:** reverting the `toDateTime64` wrap makes both `test_dollar_kline_export_preserves_clickhouse_timestamp_scale` and `test_dollar_month_rollup_matches_hf_day_scoped` fail; restoring it makes them pass.
- New `test_dollar_kline_export_preserves_clickhouse_timestamp_scale` runs the real arrow path and asserts the 2024 dates survive (not merely "not epoch").
- Tightened `test_dollar_month_rollup_matches_hf_day_scoped` from "drop datetimes" to full-frame equality now that both paths agree.
- Audited every other `Datetime` cast site (time export, both rollups, `asset_insert_to_tdw`) — the dollar export was the only bug.

## Deploy / remediation
No schema or job change. After deploy, the six live dollar files keep their 1970 timestamps only until the next publish — the daily dollar jobs fully replace each file via `delete_patterns`, so they self-heal on the next ~04:01 UTC run (or a manual Dagit trigger). No backfill needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v1.16.1 on June 4, 2026
+- Fix Binance spot dollar-kline Hugging Face exports collapsing every timestamp to ~1970 under polars >=1.40 by emitting millisecond-precision DateTime64 so the Arrow round-trip preserves the real dates.
+
 # v1.16.0 on June 4, 2026
 - Add a 10-minute stateless job mirroring the 12 Binance spot kline series to monthly Parquet files on a local mount.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "quickstart_etl"
-version = "1.16.0"
+version = "1.16.1"
 description = "Trades Warehouse Control Plane"
 readme = "README.md"
 requires-python = ">=3.11"
--- a/tdw_control_plane/utils/publish_binance_spot_dollar_kline_snapshot_to_huggingface.py
+++ b/tdw_control_plane/utils/publish_binance_spot_dollar_kline_snapshot_to_huggingface.py
@@ -173,8 +173,12 @@ def _get_binance_spot_dollar_klines(
         arrow_table = client.query_arrow(
             f"""
             SELECT
-                min(source_start_datetime) AS start_datetime,
-                max(source_end_datetime) AS end_datetime,
+                -- toDateTime64(_, 3): emit ms-precision so polars >=1.40 from_arrow keeps the
+                -- value. A plain min()/max() on these DateTime columns arrives as epoch seconds
+                -- and the Datetime("ms") cast below would reinterpret it as ms, collapsing every
+                -- timestamp to ~1970 (the same guard dollar_month uses in binance_spot_kline_rollups).
+                toDateTime64(min(source_start_datetime), 3) AS start_datetime,
+                toDateTime64(max(source_end_datetime), 3) AS end_datetime,
                 target_dollar_bar_id AS dollar_bar_id,
                 argMin(source_open, tuple(source_start_datetime, source_dollar_bar_id)) AS open,
                 max(source_high) AS high,
--- a/tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py
+++ b/tests/origo_source_native/test_publish_binance_spot_dollar_klines_to_huggingface.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import importlib
 import json
 from collections.abc import Callable
-from datetime import datetime
+from datetime import UTC, datetime
 from pathlib import Path
 
 import polars as pl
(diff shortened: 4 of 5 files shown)


zero-bang (approved):
LGTM — correct, minimal, and well-tested.

**Fix**: Wrapping the timestamp aggregates in `toDateTime64(_, 3)` makes ClickHouse emit ms precision over the wire, so the downstream `.cast(Datetime("ms", "UTC"))` is now value-preserving instead of reinterpreting epoch-seconds as milliseconds (the polars >=1.40 `from_arrow` no-rescale change). This exactly matches the `dollar_month` twin in `binance_spot_kline_rollups.py` (lines 278-279), which is the right call for consistency — preferable to diverging toward the time export's `(col.cast(Int64) * 1000)` compensation.

**Verified against the tree**:
- The dollar export was indeed the only site doing a plain `min()`/`max()` + `cast(Datetime("ms"))` with no scale compensation.
- The new `test_dollar_kline_export_preserves_clickhouse_timestamp_scale` exercises the real arrow path (not a mock) and asserts the 2024 day survives, not merely "not epoch" — a meaningful regression guard.
- Tightening `test_dollar_month_rollup_matches_hf_day_scoped` to full-frame equality is correct now that both paths share timestamp handling. The remaining `_DATETIME_COLUMNS` drop at `test_publish_binance_spot_klines_to_mount.py:311` is in a different test (`test_dollar_open_day_from_raw_matches_base_refresh`), where the open-day recompute can legitimately differ on datetime boundaries — correctly left as-is with its own range assertions.
- CHANGELOG ordering (newest-at-top) matches repo convention; version bump 1.16.0 -> 1.16.1 is appropriate.

**Forward-looking**: There are now two compensation idioms in the codebase for the same polars >=1.40 second-precision quirk — `toDateTime64(_, 3)` (rollups + this export) and `*1000` in the time export. A future export that copies neither and does a bare `cast(Datetime("ms"))` on a ClickHouse `DateTime` column will silently reintroduce the 1970 collapse, and nothing gates against it. Worth keeping in mind if more HF exports get added — a shared helper or a precision assertion on the arrow schema could make the invariant explicit. Not blocking.

PR #236, opened 2026-06-05
feat(origo): versioned mmap-ready Arrow bar store from the parquet mirror

mikkokotila:
Closes #237

## Summary
Tier-1 of the bar-store stack: a Dagster layer that projects the 12-series Parquet mirror (#232) into a **versioned, mmap-ready Arrow** store under `LOCAL_ARROW_DIR` (default `/opt/arrow`), for a downstream consumer library + HTTP service (both out of scope here) to `mmap` and `searchsorted` over a zero-copy `ts` view.

### What's included
- **`assets/build_bar_store_arrow.py`** — one partitioned asset, 12 partitions (one per mirror series). Per series: read the whole series Parquet, derive `ts` (Int64 ns UTC; `datetime` for time bars, `end_datetime` for dollar), sort + dedupe so `ts` is strictly increasing (searchsorted contract), carry every measure column **verbatim at full precision — no downcast**, combine to a single record batch (`rechunk`), write **uncompressed** Arrow IPC (Feather v2). Publish is atomic, never in place: temp in same dir → fsync → `os.replace` to `<series>.<ver>.arrow`, then an atomic relative `latest.arrow` symlink swap **last**. **Version = content hash** (byte-identical rebuild is a no-op; a pinned version is reproducible). A per-series `flock` brackets check-publish-flip-reap; a **monotonic freshness guard** stops a slow out-of-order run flipping `latest` back to staler data; retention keeps the last `N>=3`, reaping older only after a ~10-min grace so in-flight `mmap` / pinned-version reads never `ENOENT` mid-swap.
- **Trigger** — `bar_store_source_sensor`, a `RunStatusSensorDefinition` that fans out one Arrow run per series (run-keyed to the mirror run) when `publish_binance_spot_klines_to_mount_job` **succeeds**. Event-driven off the mirror's completion, not a clock or a file poll.
- **Mirror cadence** — `publish_binance_spot_klines_to_mount_schedule` now runs **every minute** (was `*/10`), so the mirror — and the Arrow store it triggers — track the 1-minute ClickHouse `_latest` projections (Arrow freshness ~1–2 min, was ~10).
- **Compose** — Compose-managed `tdw-arrow` volume (`tdw-control-plane_arrow`) on the `dagster` service (prod) + `./arrow_data` dev bind mount; `arrow_data/` gitignored. Mirrors the `tdw-parquet` pattern; no manual host step.

### Reproducibility (the no-downcast decision)
The store is a **re-layout** of the mirror, not a lossy projection: measure columns are byte-identical (Float64/Int64), and `ts`/`start_ts` are an exact ms→ns re-encode. A wider system reads multiple stores and none may silently diverge, so f32/i32 narrowing was deliberately dropped. `test_store_values_are_bit_identical_to_parquet` locks it in.

## Validation (local; ephemeral ClickHouse via the conftest harness, polars 1.41.2)
- `pytest tests/origo_source_native -q` → **112 passed** (incl. 11 new tests: shaping/sort/dedupe, dollar ts=end, **bit-identical-vs-parquet**, single-batch mmap + zero-copy ts, atomic swap keeps a pre-opened handle readable, content-hash skip/reproducibility, monotonic guard, retention+grace, run-request fan-out, sensor↔mirror-job wiring).
- pyright (strict): **error-count delta 0** vs `origin/main` (1212→1212); the new module has 0 errors; `filesAnalyzed` ratchets up.
- ruff: clean on the new module.

## Acceptance criteria (per series) — all covered
- `read_ipc(latest, memory_map=True, rechunk=False).n_chunks() == 1`
- `df["ts"].to_numpy(allow_copy=False)` succeeds; `ts` strictly increasing
- a handle opened **before** a publish still reads the old version **after** the swap

## Deploy
`docker compose up` auto-creates the Compose-managed `tdw-control-plane_arrow` volume (host `/var/lib/docker/volumes/tdw-control-plane_arrow/_data`, container `/opt/arrow`) — no manual host step. The per-minute mirror schedule drives the run-status sensor, which fans out the 12 Arrow series on each mirror success; the store self-fills on the first successful mirror run. Dagster is the sole writer; consumers mount the volume read-only.

## Operational note
At 1-min cadence × 12 partitions, every mirror success fans out ~12 Arrow runs/min (~17k/day, plus ~1.4k mirror runs/day). With `DefaultRunCoordinator` and no run retention, Dagster run-storage will grow — worth a run-retention policy, or (if run sprawl bites) collapsing the 12 partitions into a single non-partitioned Arrow run (1/min) at the cost of per-series independence. Left partitioned per the original spec.

### DIFF
--- a/tdw_control_plane/assets/build_bar_store_arrow.py
+++ b/tdw_control_plane/assets/build_bar_store_arrow.py
@@ -1,0 +1,371 @@
+"""Project the 12 Parquet mirror series into a versioned, mmap-ready Arrow bar store.
+
+The Parquet mirror (``publish_binance_spot_klines_to_mount``) keeps a tree of
+monthly files under ``LOCAL_PARQUET_DIR``. This asset reads one whole series,
+shapes it into a single uncompressed Arrow IPC (Feather v2) record batch, and
+publishes it under ``LOCAL_ARROW_DIR`` so a consumer can ``mmap`` the file and
+``searchsorted`` over a zero-copy ``ts`` view. The shaping contract is:
+
+1. read every monthly Parquet file for the series;
+2. sort ascending by ``ts`` (Int64 nanoseconds, UTC) and drop duplicate ``ts``
+   so the index is strictly increasing (the searchsorted correctness contract);
+3. carry every measure column verbatim at full precision -- no downcast. The store
+   is a re-layout of the mirror, not a lossy projection, so it stays bit-for-bit
+   reproducible against the Parquet (and thus ClickHouse / Hugging Face). The only
+   derived columns are ``ts`` / ``start_ts``, an exact ms->ns re-encode of the bar
+   datetimes;
+4. combine to a single record batch (``rechunk``) so the consumer sees one chunk;
+5. write Feather v2 / Arrow IPC, uncompressed (mmap needs raw bytes).
+
+Publishing is atomic and never in place: the bytes land in a same-dir temp file
+that is fsynced and ``os.replace``-d into ``<series>.<version>.arrow``, then a
+fresh relative symlink is ``os.replace``-d onto ``latest.arrow`` last. The
+version is the content hash, so a byte-identical rebuild is a no-op (no churn)
+and a pinned version is reproducible. A per-series lock serializes the
+check-publish-flip-reap critical section, a monotonic freshness guard keeps a
+slow out-of-order run from flipping ``latest`` back to staler data, and retention
+keeps the last ``RETENTION_KEEP`` versions, reaping older ones only after a grace
+window so in-flight ``mmap`` and pinned-version reads never ``ENOENT`` mid-swap.
+
+Dagster (this asset) is the sole writer to ``LOCAL_ARROW_DIR``; consumers
+bind-mount it read-only.
+"""
+
+import fcntl
+import hashlib
+import io
+import os
+import time
+import uuid
+from dataclasses import dataclass
+from pathlib import Path
+
+import polars as pl
+from dagster import AssetExecutionContext, Config, RunRequest, StaticPartitionsDefinition, asset
+
+from tdw_control_plane.assets.publish_binance_spot_klines_to_mount import (
+    DEFAULT_MOUNT_DIR,
+    SPECS,
+    MountKlineSpec,
+)
+
+DEFAULT_ARROW_DIR = "/opt/arrow"
+LATEST_NAME = "latest.arrow"
+# How many published versions to retain per series, and how long a superseded
+# version must linger before it can be reaped. >= 3 versions and a ~10-minute
+# grace keep in-flight mmap / pinned-version reads from racing a reap.
+RETENTION_KEEP = 3
+REAP_GRACE_SECONDS = 600
+ORPHAN_TMP_MAX_AGE_SECONDS = 3600
+# Truncated SHA-256 of the IPC payload; 16 hex chars (64 bits) is collision-safe
+# for a per-series version line that turns over a few times an hour.
+VERSION_HEX = 16
+
+# Measure columns shared by both families, carried verbatim from the Parquet mirror
+# at their native dtype (Float64 / Int64) -- no downcast, so the store is bit-for-bit
+# reproducible against the mirror. Canonical order matches the mirror / HF exports.
+_VALUE_COLUMNS: tuple[str, ...] = (
+    "open",
+    "high",
+    "low",
+    "close",
+    "mean",
+    "std",
+    "volume",
+    "maker_ratio",
+    "no_of_trades",
+    "open_liquidity",
+    "high_liquidity",
+    "low_liquidity",
+    "close_liquidity",
+    "liquidity_sum",
+    "maker_volume",
+    "maker_liquidity",
+)
+# Output column order. ``ts`` is the canonical index (datetime for time bars,
+# end_datetime for dollar bars); dollar bars also carry their open time and id.
+_TIME_OUTPUT_COLUMNS: tuple[str, ...] = ("ts", *_VALUE_COLUMNS)
+_DOLLAR_OUTPUT_COLUMNS: tuple[str, ...] = ("ts", "start_ts", "dollar_bar_id", *_VALUE_COLUMNS)
+
+
+@dataclass(frozen=True)
+class BarSeriesBuild:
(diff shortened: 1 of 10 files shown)


tdw_control_plane/assets/build_bar_store_arrow.py
```diff
@@ -0,0 +1,371 @@
...
+    ``ts`` is Int64 nanoseconds (UTC): the bar ``datetime`` for time series, the
+    bar ``end_datetime`` for dollar series. Measure columns are carried verbatim
+    (no downcast) so the frame is bit-for-bit reproducible against the Parquet.
+    Duplicate ``ts`` rows (rare; the mirror is already grouped) keep the last
+    occurrence so the index is strictly increasing for searchsorted."""
+    files = series_source_files(spec, parquet_root)
+    if not files:
+        return BarSeriesBuild(pl.DataFrame(), 0, 0)
+
+    raw = pl.read_parquet([str(path) for path in files])
```
zero-bang, line 158: Engineering concern worth recording before prod (not blocking): every sensor tick re-reads the **entire series history** here (`glob("**/*.parquet")` over all months back to 2020-01) and `publish_series` then rewrites a **full uncompressed Arrow file**. For `time_1m` that's ~6 years of 1-minute bars (~3M+ rows) read, sorted, hashed, and re-serialized to a multi-hundred-MB file — every minute, for all 12 series.

The sharper edge is interaction with versioning + grace: because the in-progress bar of *every* series changes each minute, the content hash changes each minute, so a brand-new `<series>.<ver>.arrow` is written every minute. `RETENTION_KEEP=3` doesn't cap this in the short term — `reap_old_versions` won't unlink anything younger than `REAP_GRACE_SECONDS=600`, so at 1 new version/min you transiently hold ~10+ full-size copies of the largest series before reaping catches up. That's both sustained write amplification (hundreds of MB/min) and a transient disk footprint several× the steady state, on a Compose-managed volume.

Forward-looking failure mode: the `tdw-control-plane_arrow` volume fills (or the per-minute build runs longer than 60s and runs start overlapping) once `time_1m` history is large enough — silently, since nothing here bounds it. Worth either capping how far back a tick re-materializes (e.g. only the open + previous month, mirroring the mirror's own `months_for_run` strategy, and append/merge) or at minimum logging the per-series file size + retained-version count so the growth is observable. Left-as-is is defensible for Tier-1, but please call out the chosen tradeoff explicitly.

tdw_control_plane/assets/build_bar_store_arrow.py
```diff
@@ -0,0 +1,371 @@
...
+        return []
+    return sorted(base.glob("**/*.parquet"))
+
+
+def _output_columns(family: str) -> tuple[str, ...]:
+    return _TIME_OUTPUT_COLUMNS if family == "time" else _DOLLAR_OUTPUT_COLUMNS
+
+
+def build_series_frame(spec: MountKlineSpec, parquet_root: Path) -> BarSeriesBuild:
+    """Read, shape, sort, dedupe, downcast, and single-batch one series.
```
zero-bang: Docstring says "Read, shape, sort, dedupe, **downcast**, and single-batch one series" — but the whole design (and the rest of this very docstring) is *no downcast*; measures are carried verbatim. The stale verb directly contradicts the bit-identical contract this PR emphasizes and will mislead a future reader. Drop `downcast` (e.g. "Read, shape, sort, dedupe, and single-batch one series").

zero-bang (commented):
Solid, careful piece of work — the atomic publish (temp → fsync → os.replace → relative-symlink swap last), the per-series flock around check-publish-flip-reap, the monotonic freshness guard, content-hash versioning, and the keep-N + grace reap all hang together correctly, and the open-handle-survives-swap guarantee holds (Linux keeps the unlinked inode alive for any open fd/mmap, so the 10-min grace is really only protecting the read-symlink→open gap). I verified the 16 measure columns + no_of_trades exist in both TIME_KLINE_COLUMNS and DOLLAR_KLINE_COLUMNS, so the verbatim select won't crash any partition. Tests are meaningful and cover the contract well.

No blocking correctness issues found. Two non-blocking notes inline: one engineering concern about per-minute full re-materialization cost (the predictable next failure mode), and one stale word in a docstring. The run-sprawl note in your description is well-taken; the disk write-amplification angle below is the part I'd want on the record before this runs in prod.

tdw_control_plane/assets/build_bar_store_arrow.py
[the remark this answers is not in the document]
mikkokotila: Fixed in d646ca3 — dropped the stale "downcast" verb; the summary line now reads "at full precision".

tdw_control_plane/assets/build_bar_store_arrow.py
```diff
@@ -0,0 +1,371 @@
...
+            existing_max = _existing_max_ts(latest) if latest.is_symlink() else None
+            if existing_max is not None and new_max is not None and new_max < existing_max:
+                status = "skipped_not_newer"
+            else:
+                _atomic_swap_symlink(latest, target.name)
+                status = "published"
+        reaped = reap_old_versions(
+            directory, series, keep=RETENTION_KEEP, grace_seconds=REAP_GRACE_SECONDS, now=now
+        )
+    return PublishOutcome(status, version, reaped)
```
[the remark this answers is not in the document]
mikkokotila: Good catch — fixed in 84293e6: the version file is now written only in the publish branch, under the per-series lock, so a skipped_not_newer / unchanged tick never leaves an unreferenced version to occupy a RETENTION_KEEP slot or churn disk.

tdw_control_plane/assets/build_bar_store_arrow.py
mikkokotila: Acknowledged and on the record — the per-minute full re-materialization and the 12-way run fan-out are in the PR description (Operational note), the slice issue #237 Risks, and project memory; mitigations are a run-retention policy or collapsing the partitions into one Arrow run. 84293e6 also drops the wasted write on skipped ticks, trimming part of the churn.

tdw_control_plane/assets/build_bar_store_arrow.py
mikkokotila: Fixed in d646ca3.

zero-bang (approved):

PR #239, opened 2026-06-05
fix(origo): write the Arrow bar store as a single record batch

mikkokotila:
Closes #238

## Summary
The Arrow bar store (#236) was writing **multiple record batches** for any series past polars' ~122k-row IPC default. A `memory_map=True` reader — the consumer path this store exists for — surfaces those batches as multiple chunks, so `df["ts"].to_numpy(allow_copy=False)` **raises**, breaking the single-batch zero-copy contract. Found live in prod: `time_1m` 28 batches, `dollar_1M` 42, `time_15m` 2.

**Fix:** `_ipc_payload` now passes `record_batch_size=max(df.height, 1)` to `write_ipc`, forcing exactly one record batch regardless of size. The bytes stay deterministic, so content-hash versioning remains reproducible.

**Why it slipped #236:** the tests used 3-row frames and one non-mmap spot-check, both below the split threshold. This PR adds `test_large_series_publishes_single_record_batch` (200k rows, read via `memory_map=True`) — A/B-proven to fail without the fix (`assert 2 == 1`).

## Validation
- `pytest tests/origo_source_native -q` → **113 passed** (incl. the new large-data test).
- pyright strict: error-count delta 0 vs main; ruff clean.

## Deploy / self-heal
No schema change. On deploy, `record_batch_size` changes the serialized bytes → a new content-hash version → each of the 12 series republishes once as a single batch on the next mirror success; old multi-batch versions are reaped after the retention grace. No backfill, no manual step.

### DIFF
--- a/tdw_control_plane/assets/build_bar_store_arrow.py
+++ b/tdw_control_plane/assets/build_bar_store_arrow.py
@@ -14,7 +14,9 @@ publishes it under ``LOCAL_ARROW_DIR`` so a consumer can ``mmap`` the file and
    reproducible against the Parquet (and thus ClickHouse / Hugging Face). The only
    derived columns are ``ts`` / ``start_ts``, an exact ms->ns re-encode of the bar
    datetimes;
-4. combine to a single record batch (``rechunk``) so the consumer sees one chunk;
+4. write one record batch -- ``record_batch_size`` forces a single batch (polars
+   otherwise splits large frames, and a ``memory_map`` reader surfaces each batch
+   as its own chunk), so the consumer sees one contiguous chunk;
 5. write Feather v2 / Arrow IPC, uncompressed (mmap needs raw bytes).
 
 Publishing is atomic and never in place: the bytes land in a same-dir temp file
--- a/tdw_control_plane/assets/build_bar_store_arrow.py
+++ b/tdw_control_plane/assets/build_bar_store_arrow.py
@@ -251,7 +251,11 @@ def reap_old_versions(
 
 def _ipc_payload(df: pl.DataFrame) -> bytes:
     sink = io.BytesIO()
-    df.write_ipc(sink, compression="uncompressed")
+    # record_batch_size >= height forces a single record batch. Without it polars
+    # splits frames past its default (~122k rows) into multiple batches, which a
+    # ``memory_map=True`` reader surfaces as multiple chunks -- breaking the single
+    # batch / zero-copy ``ts`` contract for every non-trivial series.
+    df.write_ipc(sink, compression="uncompressed", record_batch_size=max(df.height, 1))
     return sink.getvalue()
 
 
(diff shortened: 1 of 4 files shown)


tdw_control_plane/assets/build_bar_store_arrow.py
```diff
@@ -251,7 +251,11 @@ def reap_old_versions(
 
 def _ipc_payload(df: pl.DataFrame) -> bytes:
     sink = io.BytesIO()
-    df.write_ipc(sink, compression="uncompressed")
+    # record_batch_size >= height forces a single record batch. Without it polars
+    # splits frames past its default (~122k rows) into multiple batches, which a
+    # ``memory_map=True`` reader surfaces as multiple chunks -- breaking the single
+    # batch / zero-copy ``ts`` contract for every non-trivial series.
+    df.write_ipc(sink, compression="uncompressed", record_batch_size=max(df.height, 1))
```
zero-bang, line 260: Non-blocking doc consistency: the module docstring (step 4, line 17) still says the single record batch comes from `rechunk` — "combine to a single record batch (``rechunk``) so the consumer sees one chunk". This PR proves that's exactly the assumption that failed: `rechunk()` controls in-memory chunking, not `write_ipc` batching. A future engineer reading that line could conclude `record_batch_size` here is redundant and remove it, reintroducing the multi-batch bug. Suggest updating step 4/5 of the docstring to attribute the single-batch guarantee to `record_batch_size` in `_ipc_payload`.

zero-bang (approved):
Solid, well-scoped fix. The root cause is real and the diagnosis is right: `write_ipc` batches independently of in-memory `rechunk()`, so a `memory_map=True` reader sees one chunk per record batch and `to_numpy(allow_copy=False)` raises on anything past polars' ~122k-row default. Forcing `record_batch_size=max(df.height, 1)` collapses to one batch, the `max(...,1)` is a sensible guard against a 0-row batch size, and the bytes stay deterministic so content-hash versioning still holds. The new 200k-row test exercises the real `build_series_frame` -> `publish_series` -> `read_ipc(memory_map=True)` path and asserts `n_chunks() == 1`, which is exactly the contract that was broken — and is below-threshold-proof against the original tests.

Good call documenting the self-heal: the byte change yields a new content-hash version per series, republished once on the next mirror success, old versions reaped after the grace window. No backfill needed.

One non-blocking doc-consistency note inline. Approving.

Engineering note: `rechunk()` at line 171 is now redundant *for write batching* (record_batch_size owns that), though it still makes the buffer contiguous before serialization — harmless to keep. Forward-looking: the next predictable failure mode is the polars version. `record_batch_size` on `write_ipc` is a relatively recent addition and polars is unpinned in requirements.txt; a future image with an older polars would raise `TypeError` here, and a newer one that changes default serialization byte layout would silently re-hash every series version. Worth a floor pin (`polars>=...`) at some point so this contract can't regress out from under the store.

tdw_control_plane/assets/build_bar_store_arrow.py
mikkokotila: Fixed in the latest commit — step 4 now credits `record_batch_size` (rechunk only consolidates in-memory chunks; the write-side batch size is what guarantees one record batch).

PR #246, opened 2026-06-05
chore: update issue templates

mikkokotila:
Closes #245

## Summary
- replace the legacy RFC issue template with a PRD issue form
- keep the slice issue form loadable by removing its empty title field
- bump version to v1.17.2 and update changelog

## Validation
- ruby YAML parse for prd.yml and slice.yml
- tools/version_gate.py
- git diff --check


zero-bang (approved):
LGTM — clean template swap.

Validated:
- `prd.yml` and `slice.yml` both parse (yaml.safe_load); PRD form has 17 body fields, slice form loses only the `title: ""` line.
- No dangling references to `RFC-Template`/`RFC-XXXX` anywhere else in the repo (the only hits were inside the deleted file), and there's no `config.yml` template chooser pointing at it.
- Version gate passes; `1.17.1 → 1.17.2` bump and CHANGELOG entry are consistent.

Two non-blocking notes:
- All 17 PRD fields are `required: true`. That's a real UX shift from the old free-form markdown template — an issue form hard-blocks submission until every textarea (Market Desirability, Drawbacks, Alternatives, …) is filled, where the markdown template let you submit with placeholders. If that strictness is the intent of converting to a form, great; just calling it out so it's a deliberate choice. Note the PRD form also drops the old `title: "RFC-XXXX: "` prefill, so filed PRDs get no default title prefix.
- I couldn't independently confirm that the empty `title: ""` key was what made the slice form fail to load (GitHub accepts an empty default title in other repos), but removing it yields a valid form either way, so it's harmless regardless of the exact root cause.

Forward-looking: the next predictable failure mode is template drift — if a future PRD form field is renamed or made optional, nothing mechanically checks it against the slice form's "Work Slices" linkage. Worth a lightweight CI yaml-lint on `.github/ISSUE_TEMPLATE/*.yml` if these forms churn.

PR #252, opened 2026-06-15
feat(origo): add spot depth200 source projection

mikkokotila:
Closes #249

Tests:
- /tmp/tdw-control-plane-pytest/bin/python -m pytest tests/origo_source_native -q --maxfail=1
- pyright --pythonpath /tmp/tdw-control-plane-pytest/bin/python tdw_control_plane/assets/create_binance_spot_depth200_snapshots_table_origo.py tdw_control_plane/assets/sync_binance_spot_depth200_snapshots_to_origo.py tdw_control_plane/assets/create_binance_spot_depth200_1m_table_origo.py tdw_control_plane/assets/refresh_binance_spot_depth200_1m_origo.py tdw_control_plane/assets/reconcile_binance_spot_depth200_partition_state_origo.py
- /tmp/tdw-control-plane-pytest/bin/python -m ruff check tools tests/tools
- /tmp/tdw-control-plane-pytest/bin/python tools/typing_gate.py --pyright-json /tmp/tdw-pyright.json --base-budget /tmp/tdw-base-typing-budget.json
- /tmp/tdw-control-plane-pytest/bin/python tools/fail_loud_gate.py --base-budget /tmp/tdw-base-fail-loud-budget.json
- /tmp/tdw-control-plane-pytest/bin/python tools/version_gate.py --pr-title "feat(origo): add spot depth200 source projection" --base-pyproject /tmp/tdw-base-pyproject.toml --head-pyproject pyproject.toml --base-changelog /tmp/tdw-base-CHANGELOG.md --head-changelog CHANGELOG.md
- /tmp/tdw-control-plane-pytest/bin/python tools/slice_gate.py --pr-title "feat(origo): add spot depth200 source projection" --pr-body-file /tmp/tdw-pr-body.md --pr-files-file /tmp/tdw-pr-files.txt --template .github/ISSUE_TEMPLATE/slice.yml --repo Vaquum/tdw-control-plane


zero-bang (approved):
LGTM. This is a faithful, well-tested clone of the depth20 source-native pipeline for the 200-level book.

**Verified:**
- All 5 new assets diff cleanly against their depth20 originals (only the two intentional divergences noted below).
- `definitions.py` has exact registration parity — every depth20 entry (asset list, schedule, the 5 jobs, and the `_bar_store_on_mirror_success` asset set) has a matching depth200 counterpart.
- The fixture is valid: all 16 NDJSON lines carry exactly 200 bids/200 asks, satisfying the `_price_levels` 200-level guard.
- Tests are actually stronger than depth20's — the new `test_..._sync_inserts_real_history_payload` and `test_..._refresh_projects_latest_synced_snapshot` materialize the real pipeline and recompute mid/spread/depth/imbalance in Python, asserting `isclose`.
- Run-key namespace `binance_spot_depth200::` is `::`-separated, so no collision with `binance_spot_depth20::` despite the shared prefix.
- All 9 CI checks pass (tests, typing, CodeQL included).

**Two non-blocking divergences from depth20 (both safe):**
1. `sync_binance_spot_depth200_snapshots_to_origo.minute_start_from_context` drops the `if value is None` guard that depth20 keeps. The typing check passing confirms `context.partition_key` is typed `str` here, so this is harmless — just a cosmetic inconsistency between the two sibling modules. Consider keeping them aligned to avoid future "which one is right?" confusion.
2. `refresh_..._1m`'s `_insert_minute_row` selects the latest snapshot with `ORDER BY source_timestamp_ms DESC` rather than depth20's `ORDER BY datetime DESC`. Since `datetime` is derived from `source_timestamp_ms` at ms precision these are equivalent, and ordering on the monotonic source timestamp (also the ReplacingMergeTree version column) is arguably the better choice.

**Forward-looking:** the most predictable failure mode is the `_price_levels` strict `len(value) != 200` check — if Binance ever returns a partial book (thin market, maintenance window, or an API that caps below 200), the whole minute hard-fails and the partition never materializes. That's a deliberate fail-loud stance consistent with depth20, but worth keeping in mind as the first thing to revisit if you see depth200 partitions stalling.

Issue #123, opened 2026-01-19
chore: remove redundant services from docker


mikkokotila:
- [x] chore: remove tabix
- [x] chore: remove metabase
- [x] chore: remove `metabase-data` volume
- [x] chore: remove `grafana-data` volume

Issue #255, opened 2026-06-15
PRD-0002: Binance spot depth Arrow snapshot store


mikkokotila:
### Objective

Expose Binance spot depth20 and depth200 raw order-book snapshots as mmap-ready Arrow IPC files under `/opt/arrow` so Arrow-aware consumers can read the same live depth datasets without querying ClickHouse.

### Summary

Add a small Dagster Arrow publisher for the existing `binance_spot_depth20_snapshots` and `binance_spot_depth200_snapshots` ClickHouse tables. The publisher writes one uncompressed single-record-batch Arrow IPC file per depth series, keeps the existing atomic `latest.arrow` symlink/version pattern, and preserves book levels as fixed-size Arrow nested arrays instead of flattening hundreds of level columns.

### Is This Really Necessary?

Yes. Depth20 and depth200 now exist in Origo, but low-latency local consumers still need a direct Arrow file path. Reusing `/opt/arrow` avoids another data plane and gives the same operational model as the current bar store.

### Named Concepts

- Depth snapshot Arrow store: the `/opt/arrow` files for raw depth20/depth200 source-native snapshots.
- Depth20 snapshot file: `/opt/arrow/depth20_snapshots/latest.arrow`.
- Depth200 snapshot file: `/opt/arrow/depth200_snapshots/latest.arrow`.
- Book side column: `bids` or `asks`, stored as `fixed_size_list<struct<price: float64, qty: float64>>[depth]`.
- Zero-copy consumer: a consumer that memory-maps Arrow IPC and reads Arrow buffers directly without materializing Python objects, pandas frames, or copied chunks.

### Outline

Deliver one implementation slice. Add one asset module that queries the two existing raw snapshot tables, shapes rows into Arrow-native depth snapshot frames, and publishes each series through the existing `/opt/arrow` atomic version/latest mechanics. Register one Dagster job with static partitions for `depth20_snapshots` and `depth200_snapshots`. Add tests proving schema, single-batch IPC, zero-copy primitive buffers, publish paths, and definitions wiring.

### Out of Scope

- Creating or changing ClickHouse depth tables.
- Changing depth20/depth200 ingestion or schedules.
- Publishing 1-minute scalar depth projections to `/opt/arrow`.
- Flattening book levels into `bid_price_001`-style columns.
- Adding Hugging Face datasets or remote publishing.
- Adding a new filesystem root beyond `/opt/arrow`.

### User Experience

[...]

PR #257, opened 2026-06-15
feat(origo): publish spot depth snapshots to Arrow

mikkokotila:
Closes #256

## What does this PR change?

Adds a manual Dagster job that publishes raw Binance spot depth20/depth200 snapshot tables into `/opt/arrow/<series>/latest.arrow` as single-record-batch, uncompressed Arrow IPC files with fixed-size nested bid/ask book columns.

PRD: #255

## Tests

- `/tmp/tdw-control-plane-pytest/bin/python -m pytest tests/origo_source_native/test_build_depth_snapshot_store_arrow.py -q`
- `/tmp/tdw-control-plane-pytest/bin/python -m pytest tests/origo_source_native -q --maxfail=1`
- `pyright --pythonpath /tmp/tdw-control-plane-pytest/bin/python tdw_control_plane/assets/build_depth_snapshot_store_arrow.py`
- `/tmp/tdw-control-plane-pytest/bin/python -m ruff check tools tests/tools`
- `/tmp/tdw-control-plane-pytest/bin/python tools/version_gate.py --pr-title "feat(origo): publish spot depth snapshots to Arrow" --base-pyproject /tmp/tdw-base-pyproject.toml --head-pyproject pyproject.toml --base-changelog /tmp/tdw-base-CHANGELOG.md --head-changelog CHANGELOG.md`
- `/tmp/tdw-control-plane-pytest/bin/python tools/slice_gate.py --pr-title "feat(origo): publish spot depth snapshots to Arrow" --pr-body-file /tmp/tdw-pr-body.md --pr-files-file /tmp/tdw-pr-files.txt --template .github/ISSUE_TEMPLATE/slice.yml --repo Vaquum/tdw-control-plane`
- `/tmp/tdw-control-plane-pytest/bin/python tools/fail_loud_gate.py --base-budget /tmp/tdw-base-fail-loud-budget.json`
- `/tmp/tdw-control-plane-pytest/bin/python tools/typing_gate.py --pyright-json /tmp/tdw-pyright.json --base-budget /tmp/tdw-base-typing-budget.json`

## LLM Review

- Codex CLI approved PRD #255 and slice #256.
- Claude CLI approved PRD #255 and slice #256.
- Codex CLI approved the staged diff.
- Claude CLI approved the staged diff.

## Checklist

- [x] I ran the project's test suite locally without errors (where applicable)
- [x] I linked the slice issue to auto-close on merge

### DIFF
--- a/tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
+++ b/tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
@@ -1,0 +1,256 @@
+from collections.abc import Sequence
+from dataclasses import dataclass
+from typing import cast
+
+import polars as pl
+import pyarrow as pa
+from dagster import AssetExecutionContext, Config, StaticPartitionsDefinition, asset
+
+from tdw_control_plane.assets.build_bar_store_arrow import BarSeriesBuild, publish_series
+from tdw_control_plane.assets.create_binance_spot_depth20_snapshots_table_origo import (
+    ClickHouseClient,
+    get_clickhouse_settings,
+    make_clickhouse_client,
+)
+from tdw_control_plane.assets.create_binance_spot_depth20_snapshots_table_origo import (
+    SNAPSHOTS_TABLE_NAME as DEPTH20_SNAPSHOTS_TABLE_NAME,
+)
+from tdw_control_plane.assets.create_binance_spot_depth200_snapshots_table_origo import (
+    SNAPSHOTS_TABLE_NAME as DEPTH200_SNAPSHOTS_TABLE_NAME,
+)
+
+DEPTH_SNAPSHOT_SERIES: tuple[str, ...] = ('depth20_snapshots', 'depth200_snapshots')
+DEPTH_SNAPSHOT_PARTITIONS = StaticPartitionsDefinition(list(DEPTH_SNAPSHOT_SERIES))
+
+BookLevel = tuple[float, float]
+
+
+class DepthSnapshotStoreConfig(Config):
+    dry_run: bool = False
+
+
+@dataclass(frozen=True)
+class DepthSnapshotSpec:
+    series: str
+    table_name: str
+    depth: int
+
+
+@dataclass(frozen=True)
+class DepthSnapshotRow:
+    ts: int
+    source_timestamp_ms: int
+    last_update_id: int
+    bids: tuple[BookLevel, ...]
+    asks: tuple[BookLevel, ...]
+
+
+_DEPTH_SNAPSHOT_SPECS: tuple[DepthSnapshotSpec, ...] = (
+    DepthSnapshotSpec('depth20_snapshots', DEPTH20_SNAPSHOTS_TABLE_NAME, 20),
+    DepthSnapshotSpec('depth200_snapshots', DEPTH200_SNAPSHOTS_TABLE_NAME, 200),
+)
+
+
+def spec_for_depth_snapshot_series(series: str) -> DepthSnapshotSpec:
+    for spec in _DEPTH_SNAPSHOT_SPECS:
+        if spec.series == series:
+            return spec
+    raise ValueError(f'Unknown depth snapshot series: {series}')
+
+
+def _snapshot_rows(result: object, spec: DepthSnapshotSpec) -> tuple[DepthSnapshotRow, ...]:
+    if not isinstance(result, Sequence) or isinstance(result, (bytes, str)):
+        raise TypeError('Expected ClickHouse row sequence for depth snapshots.')
+    if not result:
+        raise RuntimeError(f'No source rows found in {spec.table_name}.')
+
+    raw_rows = cast(Sequence[object], result)
+    rows: list[DepthSnapshotRow] = []
+    for index, raw_row in enumerate(raw_rows):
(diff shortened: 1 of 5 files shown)


tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
```diff
@@ -0,0 +1,256 @@
...
+def _book_levels(
+    value: object,
+    side: str,
+    depth: int,
+    ts: int,
+) -> tuple[BookLevel, ...]:
+    if not isinstance(value, Sequence) or isinstance(value, (bytes, str)):
+        raise TypeError(f'Expected {side} levels at ts={ts} to be a sequence.')
+    raw_levels = cast(Sequence[object], value)
+    if len(raw_levels) != depth:
```
zero-bang, line 113: Worth confirming: this requires *every* row to carry exactly `depth` levels, and a single non-conforming row aborts the whole partition publish (it propagates as `RuntimeError` out of `build_depth_snapshot_frame`, so no `latest.arrow` is flipped for that series at all). For depth20 the partial-book stream guarantees 20 levels, so this is fine. For depth200 — if the source is ever a REST `/depth?limit=200` response — a thin or halted book can legitimately return fewer than 200 levels, which would wedge `depth200_snapshots` until the book deepens.

If the upstream ingestion already guarantees exactly 200 rows per snapshot (worth verifying against the depth200 sync/reconcile path), this strict check is correct fail-loud behavior and no change is needed. If not, consider whether a short book should pad/skip the row rather than block the series. Not blocking — just flagging the assumption so it's a deliberate choice.

zero-bang (commented):
## Review: solid, with one stale-CI item to clear before merge

The asset is well-built and well-tested. It cleanly reuses the proven `build_bar_store_arrow` publish machinery (`BarSeriesBuild` + `publish_series`): atomic symlink flip, content-hash versions, monotonic freshness guard, and retention all come for free, and the `ts` Int64-ns index keeps the zero-copy `mmap`/`searchsorted` contract intact. I verified the query against both ClickHouse table schemas (`create_binance_spot_depth{20,200}_snapshots_table_origo.py`): the `(datetime, source_timestamp_ms, last_update_id, bids, asks)` projection and the `Array(Tuple(Float64, Float64))` book shape match, and `FINAL` + the dict-dedupe keep `ts` strictly increasing. The fixed-size nested book layout round-trips through polars `write_ipc` and back to a `FixedSizeListArray` of `Struct{price,qty}` — the end-to-end `mmap` test proves it. The 200k-row single-record-batch test correctly guards the one-chunk invariant. No correctness bugs found.

**Engineering note (not a defect):** reusing `BarSeriesBuild`/`publish_series` is the right call — the depth store inherits the entire atomicity/freshness/retention story verbatim instead of reimplementing it. The only divergence from the sibling asset is intentional and tested: empty source hard-fails here (`RuntimeError`) rather than the bar store's graceful `no_source` return. For a manual job that's defensible fail-loud behavior.

### CI disposition
- **`pr_checks_slice` (red) — stale, not a code/body defect.** It ran against an earlier body revision that still carried the template's `Fixes #123` example, so the gate counted two closing refs (`#256`, `#123`). The current body has exactly one (`Closes #256`) and the example line is gone, so a re-run passes. **Action: re-trigger the slice check** (e.g. an empty-commit/rebuild) so it reflects the current body.
- **`pr_checks_tests`, `pr_checks_typing`, `pr_checks_version`, CodeQL — pending.** Must be green before merge; nothing actionable yet.
- `pr_checks_fail_loud`, `pr_checks_ruleset`, `pr_checks_cc`, `pr_checks_lint` — pass.

Holding at COMMENT rather than approving only because the slice gate is red (stale) and the test/typing/version checks haven't reported yet. Once those are green, this is good to merge. One forward-looking note inline.

**Next predictable failure mode:** a depth200 snapshot row that the upstream ingestion stored with fewer than 200 levels (thin/halted book, or a truncated REST `/depth` response) would make `_book_levels` raise and fail the *entire* `depth200_snapshots` partition publish on every run until the book deepens — see inline.

tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
mikkokotila: Verified against the current depth200 ingestion path: `sync_binance_spot_depth200_snapshots_to_origo._price_levels()` requires exactly 200 bid levels and exactly 200 ask levels before inserting into ClickHouse, and `test_binance_spot_depth200_sync_inserts_real_history_payload` asserts `min(length(...)) = max(length(...)) = 200` for both sides. So the Arrow publisher is not introducing the first strictness boundary; it is preserving the existing source-table invariant and failing loud if the table is populated out-of-band or corrupted. No code change needed.

zero-bang: Confirmed — thanks for tracing the depth200 ingestion path. Since `_price_levels()` enforces exactly 200 levels per side before insert (with the `min/max length = 200` test guarding it), the publisher is preserving the existing source-table invariant rather than introducing a new boundary. The strict check is correct fail-loud behavior. Resolved, no change needed.

zero-bang (approved):

PR #262, opened 2026-06-16
fix(origo): reconcile spot depth live gaps

mikkokotila:
Closes #261

## Summary
- Adds bounded source-aware reconciliation for depth20 and depth200 raw gaps.
- Adds projection-only repair schedules for raw-present projection gaps.
- Adds Arrow repair schedules for raw-present Arrow chunk/latest gaps.
- Keeps Arrow latest.json monotonic during backfills.

## Validation
- `UV_PROJECT_ENVIRONMENT=/tmp/tdw-control-plane-py311-venv uv run --python 3.11 --extra dev python -m pytest tests/origo_source_native -q --maxfail=1` -> 142 passed.
- `UV_PROJECT_ENVIRONMENT=/tmp/tdw-control-plane-py311-venv uv run --python 3.11 --with ruff==0.15.11 ruff check tools tests/tools` -> passed.
- Focused depth/Arrow tests -> 37 passed.
- Pyright baseline total remained 1212; new touched ranges clean.

### DIFF
--- a/tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
+++ b/tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
@@ -297,6 +297,17 @@ def _atomic_write_bytes(target: Path, payload: bytes) -> None:
     os.replace(tmp, target)
 
 
+def _latest_manifest_minute(manifest_path: Path) -> datetime | None:
+    if not manifest_path.exists():
+        return None
+
+    manifest = json.loads(manifest_path.read_text(encoding='utf-8'))
+    source_partition_key = manifest.get('source_partition_key')
+    if not isinstance(source_partition_key, str):
+        raise RuntimeError(f'{manifest_path} does not contain source_partition_key.')
+    return minute_start_from_partition_key(source_partition_key)
+
+
 def publish_depth_snapshot_chunk(
     series: str,
     source_partition_key: str,
@@ -309,6 +320,11 @@ def publish_depth_snapshot_chunk(
     directory = series_store_dir(series)
     chunk = directory / relative_chunk
     _atomic_write_bytes(chunk, payload)
+    manifest_path = directory / LATEST_MANIFEST_NAME
+
+    latest_minute = _latest_manifest_minute(manifest_path)
+    if latest_minute is not None and minute_start < latest_minute:
+        return DepthSnapshotChunkPublish('skipped_not_newer', version, relative_chunk.as_posix())
 
     manifest = {
         'series': series,
@@ -321,7 +337,7 @@ def publish_depth_snapshot_chunk(
         'updated_at_unix_ns': time.time_ns(),
     }
     _atomic_write_bytes(
-        directory / LATEST_MANIFEST_NAME,
+        manifest_path,
         json.dumps(manifest, sort_keys=True).encode('utf-8') + b'\n',
     )
     return DepthSnapshotChunkPublish('published', version, relative_chunk.as_posix())
--- a/tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
+++ b/tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
@@ -1,3 +1,4 @@
+import fcntl
 import hashlib
 import io
 import json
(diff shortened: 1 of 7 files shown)


tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
```diff
@@ -309,6 +320,11 @@ def publish_depth_snapshot_chunk(
     directory = series_store_dir(series)
     chunk = directory / relative_chunk
     _atomic_write_bytes(chunk, payload)
+    manifest_path = directory / LATEST_MANIFEST_NAME
+
+    latest_minute = _latest_manifest_minute(manifest_path)
+    if latest_minute is not None and minute_start < latest_minute:
```
zero-bang: TOCTOU race on `latest.json` under concurrent builds. This reads the manifest minute, then later writes the manifest, with no lock between. The new Arrow repair schedule emits multiple run requests per tick (one per gap minute) all with `partition_key=spec.series`, and `dagster.yaml` configures no run-concurrency limit (default coordinator) — so builds for minute N and the current-newest minute can run concurrently. Interleaving: build(N=10) reads manifest=5; build(N=12) reads manifest=5; build(12) writes 12; build(10) writes 10 → `latest` regresses to 10 below the newest chunk. Chunks are written atomically so no chunk is lost, and the next repair tick detects `latest_manifest_minute(10) < 12` and republishes, so it self-heals within ~1 minute. Still, consider serializing per series — e.g. a `tag_concurrency_limits` / op concurrency key on `build_depth_snapshot_store_arrow_job` keyed by series — or guarding the manifest read+write with a file lock, so `latest` never momentarily points below the newest published chunk.

tdw_control_plane/definitions.py
```diff
@@ -696,6 +766,225 @@ def _last_completed_minute(scheduled_time: datetime | None) -> datetime:
...
+) -> list[RunRequest] | SkipReason:
+    settings = get_depth_clickhouse_settings()
+    client = make_depth_clickhouse_client(settings)
+    run_key_suffix = _scheduled_run_key_suffix(context)
+    run_requests: list[RunRequest] = []
+
+    try:
+        for minute_start, partition_key in _depth_candidate_minutes(context, spec):
+            status = _depth_store_status(client, settings.database, spec, minute_start)
+            if status.snapshot_rows == 0 and _depth_source_has_rows(spec, minute_start):
```
zero-bang, line 915: A transient upstream error here fails the entire schedule tick. `_depth_source_has_rows` (line 789) calls `response.raise_for_status()` with no `except`, so a 5xx/timeout from `/history` propagates out of this loop and errors the whole reconciliation tick — every candidate minute, not just the unreachable one. The established pattern in this file is `_binance_archive_available` (line 745), which catches `requests.RequestException` and returns `False`. Recommend mirroring it: on a request error treat the minute as 'not confirmed available' (return `False`) so it's simply retried next tick while still inside the lookback window, rather than turning a recoverable source hiccup into a failed tick.

zero-bang (commented):
Solid, well-tested change. The split into source / projection / Arrow reconciliation specs is clean, the bounded lookback is sensible, the run-key namespacing (`:source:` / `:projection:` / `:arrow:`) prevents collisions, and the monotonic `latest.json` guard correctly writes the chunk before deciding whether to advance the pointer (so backfilled older chunks are persisted but never regress `latest`). Test coverage is genuinely meaningful — it exercises ordering, the completed-minute skip, and the projection-only job graph.

Two non-blocking findings inline; neither corrupts data:

1. `latest.json` read-modify-write is now subject to a TOCTOU race because this PR newly fans out multiple concurrent Arrow builds for the same series. Self-healing, but worth serializing.
2. `_depth_source_has_rows` raises on a transient upstream error, failing the whole schedule tick — unlike the existing `_binance_archive_available`, which swallows `requests.RequestException`.

Engineering note: each of the 6 new per-minute schedules opens its own ClickHouse client and runs `_depth_store_status` over all 15 lookback minutes, and that helper always issues both the snapshot-count and projection-count `FINAL` queries even when the caller only needs one (e.g. the source schedule ignores `projection_rows`/arrow status). That's ~6×15×2 `FINAL` counts/minute. Tables are tiny so it's fine today, but consider computing only the fields each schedule consumes.

Forward-looking failure mode: if the upstream `/history` endpoint degrades (5xx/timeout) during an outage window, finding #2 means every source reconciliation tick errors for as long as the gap minute stays in the lookback, turning a recoverable source hiccup into ~15 minutes of red schedule ticks.

Not approving outright only because `pr_checks_tests`, `pr_checks_typing`, and CodeQL were still pending at review time — the change looks approvable once those go green.

tdw_control_plane/definitions.py
zero-bang: Still not addressed. `_depth_source_has_rows` continues to call `response.raise_for_status()` (line 800) with no `except requests.RequestException` guard, so a 5xx/timeout from `/history` still propagates out and errors the entire reconciliation tick for every candidate minute, not just the unreachable one. Please mirror `_binance_archive_available` and return `False` on `requests.RequestException` so the minute is simply retried next tick within the lookback window.

tdw_control_plane/assets/build_depth_snapshot_store_arrow.py
zero-bang: Still not addressed. The manifest read-modify-write in `publish_depth_snapshot_chunk` remains unguarded — no `tag_concurrency_limits` / op concurrency key on the Arrow job and no file lock around the `latest.json` read+write (this file imports neither `fcntl` nor any lock, unlike `build_bar_store_arrow.py`). With the new Arrow repair schedule fanning out multiple concurrent builds per series, `latest` can still momentarily regress below the newest published chunk. It self-heals, but please serialize per series or guard the manifest read+write with a file lock.

zero-bang (changes requested):
Some items from my previous review still need to be addressed. See my replies on the relevant threads.

zero-bang (approved):

PR #268, opened 2026-07-04
refactor(origo): rebrand repo metadata to Origo and pin sensor defaults

mikkokotila:
**NOTE:** The author must always **first review the PR themselves**, before requesting review from others, that means carefully having reviewed the full diff in GitHub.

## What does this PR change?

Pass 1 of the Origo rename (PRD #266, slice #267): rebrands repo metadata to Origo (dist name `quickstart_etl` → `origo`, description, README, slice-template prose, dev image tag `trades-warehouse-dagster` → `origo-dagster`, test fixtures `tests/fixtures/tdw/` → `origo/`, test container prefix), removes the never-existing `quickstart_etl_tests` ghost path from all 8 files that referenced it (pyproject ×4, lint contract test, both budget JSONs, both gate defaults), deletes vestigial `dagster_cloud.yaml`, and pins `default_status=RUNNING` on all 12 HuggingFace asset sensors — today they run only via manually-toggled instance state; this makes the pass-2 package rename unable to silently stop them. A new regression test asserts every sensor in `defs` defaults to RUNNING. No production-facing names change: server volumes, compose project, workflows, and the `tdw_control_plane` package itself are untouched (pass 2/3). Version 2.0.0 → 2.0.1.

## Checklist

- [x] I have reviewed full diff in “Files changed”
- [x] I left no unnecessary files in the changes
- [x] I ran the project's test suite locally without errors (where applicable) — `pytest tests/origo_source_native -q --maxfail=1`: 148 passed; `pytest tests/tools/test_lint_ci_contract.py`: 7 passed; `ruff check tools tests/tools`: clean
- [x] I updated any relevant documentation (if behavior/API/config/user/etc changed)
- [x] I added and/or updated docstrings as per [Writing Docstrings](https://github.com/Vaquum/Limen/blob/main/docs/Developer/Writing-Docstrings.md) (for any changed public functions/classes)
- [x] I updated CHANGELOG.md (unless only docs or other non-code aspect was changed)
- [x] I updated pyproject.toml (unless only docs or other non-code aspect was changed)
- [x] I added and/or updated tests (if behavior changed or new code paths added)
- [x] I validated changes manually
- [x] I validated changes with LLM
- [x] I removed any extraneous examples/comments
- [x] I linked issue to auto-close on merge when applicable

Closes #267

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #270, opened 2026-07-04
refactor(origo)!: rename tdw_control_plane package to origo

mikkokotila:

## What does this PR change?

Pass 2 of the Origo rename (PRD #266, slice #269): renames the Python package `tdw_control_plane/` → `origo/` and moves every consumer in the same commit — all 122 module-path references (self-imports, test imports, conftest reload strings), `workspace.yaml`, pyproject (`packages.find`, `[tool.dagster]` module + code-location, pyright include, isort first-party), both gate package roots (`typing_gate.py` `REQUIRED_PYRIGHT_INCLUDE` + defaults, `fail_loud_gate.py`), both budget JSONs' `package_root`, the deploy smoke-test import, and the typing-workflow contract comment. The tombstone test's guards now watch `origo.*` and a new `test_legacy_package_is_absent` fails any PR that reintroduces `tdw_control_plane`. Version 2.0.1 → 3.0.0 (breaking: import namespace).

Dagster note: the code location becomes `origo.definitions`, so schedule/sensor instigator state restarts fresh on deploy — safe by design since #268 pinned `default_status=RUNNING` on every sensor (schedules already had it). Cursors/run-key resets re-trigger only idempotent work. `location_name` is intentionally not pinned to the legacy name. Old-location entries in dagit become unloadable stubs (cosmetic).

Ruff statistics for the package are byte-identical before/after (53 pre-existing findings, not gated by law 6) — no drive-by cleanup.

## Checklist

- [x] I ran the project's test suite locally without errors (where applicable) — `pytest tests/origo_source_native -q --maxfail=1`: 149 passed; `ruff check tools tests/tools`: clean

Closes #269


.github/typing_budget.json
```diff
@@ -1,6 +1,6 @@
 {
   "schema_version": 2,
-  "package_root": "tdw_control_plane",
+  "package_root": "origo",
   "excludes": [
```
zero-bang, line 4: **Blocking — this trips the typing gate's `budget-source-ratchet`.** `tools/typing_gate.py:635` compares base-ref vs head `package_root` and fails on any change: *"package_root changed from 'tdw_control_plane' (base) to 'origo' (head). The scan surface cannot be narrowed by the PR it gates."* The gate can't tell a rename from a surface-narrowing, and the `--bootstrap` escape only applies when the base ref lacks the budget entirely (it doesn't).

The cleanest fix is to make the ratchet rename-aware: permit `base_root != head_root` **only when the surface size is preserved** — i.e. the file count under the new root equals the count under the old root and every category/pattern total is unchanged. That keeps the anti-narrowing guarantee intact while allowing a pure top-level rename. This touches `tools/typing_gate.py` (`gate_budget_source`, ~line 635) and its twin in `tools/fail_loud_gate.py:317`, so bundle both gate edits with this rename. Whichever path you take, please confirm both gates go green before re-requesting review.

.github/fail_loud_budget.json
```diff
@@ -1,6 +1,6 @@
 {
   "schema_version": 1,
-  "package_root": "tdw_control_plane",
+  "package_root": "origo",
   "excludes": [
```
zero-bang, line 4: Same blocking ratchet as `typing_budget.json` — `tools/fail_loud_gate.py:317` rejects the `package_root` change: *"The scan surface cannot be narrowed in the same PR that gates."* `pr_checks_fail_loud` fails here. Apply the same rename-aware exception described on the typing budget so this gate passes too.

zero-bang (changes requested):
The rename itself is executed cleanly and completely: I verified only two `tdw_control_plane` references remain (both intentional — the CHANGELOG entry and the new `test_legacy_package_is_absent` tombstone), the `origo/` tree carries every submodule, and pyproject/workspace/gates/budgets/deploy-smoke-test are all moved consistently. The remaining `tdw-control-plane` infra names (docker image tags, named volumes, repo URLs) are correctly left untouched — renaming a named volume would drop data, and they're out of scope for a Python-namespace pass.

However, this PR **cannot merge as-is**: two required gates (`pr_checks_typing`, `pr_checks_fail_loud`) fail deterministically and by design. Both gates run a `budget-source-ratchet` that hard-rejects any change to `package_root` in the same PR it gates (`tools/typing_gate.py:635`, `tools/fail_loud_gate.py:317`), because it cannot distinguish a legitimate rename from a malicious surface-narrowing (point `package_root` at an empty subtree → findings vanish → gate trivially passes). The `--bootstrap` escape only fires when the base ref has no budget at all; here the base has both budgets under `tdw_control_plane`, so the gates run in `compare` mode and the ratchet trips.

The two budget `package_root` edits are the exact lines that trip it — see inline comments for the resolution path. Requesting changes on the merge-blocking gate failures; the code changes themselves look correct. (CodeQL and `pr_checks_tests` are still pending; author reports 149 local passes.)

.github/typing_budget.json
mikkokotila: Resolved via #272 (slice #271, merged as v2.1.0) and this branch's merge of main: the ratchet now permits a root change only when the base root directory is absent from the head tree AND every ratchet total is identical to base (totals-neutral rename — your refinement from the #272 review). This PR satisfies both conditions: `tdw_control_plane/` is gone from the head tree and both budgets change only the `package_root` string. `pr_checks_typing` is green on the current head.

.github/fail_loud_budget.json
mikkokotila: Resolved via #272 / merge of main, same as the typing budget — `pr_checks_fail_loud` is green on the current head (also verified locally against the base-ref budget: FAIL-LOUD GATE -- PASS).

zero-bang (approved):

PR #272, opened 2026-07-04
feat(gates): allow package_root rename when the base root is absent

mikkokotila:

## What does this PR change?

Slice #271 (unblocks #269 / PR #270): the typing and fail-loud budget-source ratchets rejected **any** `package_root` change, which also rejected true package renames — pass 2 of the Origo rename failed both gates. A root change is now legal **only when the base root directory no longer exists in the head tree AND every ratchet total is identical to base** (a totals-neutral rename); pointing the gate at a subtree while the old root still exists stays blocked, and no ratchet total can move in the same PR as a root change (review finding by zero-bang — the earlier old-root-absence-only rule left a mv-plus-clean-subtree channel). Residual risk stated plainly: a rename-shaped PR can still relocate files outside the scan surface, exactly as any PR always could by moving files out of the root; operator review is the backstop. Six regression tests cover allow/block/totals-neutrality for both gates, and `pr_checks_ruleset` now runs the new test file. Version 2.0.1 → 2.1.0.

## Checklist

- [x] I ran the project's test suite locally without errors (where applicable) — `pytest tests/tools -q`: 34 passed; `ruff check tools tests/tools`: clean

Closes #271

### DIFF
--- a/tests/tools/test_budget_source_ratchet.py
+++ b/tests/tools/test_budget_source_ratchet.py
@@ -14,7 +14,7 @@ TYPING_BUDGET: Final[dict[str, object]] = {
     'package_root': 'tools',
     'excludes': [],
     'patterns': {},
-    'pyright_errors': {'total': 0},
+    'pyright_errors': {'total': 5},
     'any_references': {'total': 0},
 }
 
@@ -22,7 +22,7 @@ FAIL_LOUD_BUDGET: Final[dict[str, object]] = {
     'schema_version': 1,
     'package_root': 'tools',
     'excludes': [],
-    'categories': {},
+    'categories': {'bare_except': {'total': 2}},
 }
 
 
@@ -42,6 +42,12 @@ def _budget(template: dict[str, object], package_root: str) -> dict[str, object]
     return budget
 
 
+def _absent_root(tmp_path: Path) -> str:
+    root = f'renamed_away_{tmp_path.name}'
+    assert not (REPO_ROOT / root).is_dir()
+    return root
+
+
 def _root_failures(failures: list[str]) -> list[str]:
     return [failure for failure in failures if 'package_root changed' in failure]
 
@@ -49,7 +55,7 @@ def _root_failures(failures: list[str]) -> list[str]:
 def test_typing_gate_allows_root_rename_when_base_root_absent(tmp_path: Path) -> None:
     typing_gate = _load('typing_gate')
     base_path = tmp_path / 'base_budget.json'
-    base_path.write_text(json.dumps(_budget(TYPING_BUDGET, 'renamed_away_pkg_xyz')))
+    base_path.write_text(json.dumps(_budget(TYPING_BUDGET, _absent_root(tmp_path))))
 
     failures = typing_gate.gate_budget_source(
         str(base_path), False, _budget(TYPING_BUDGET, 'tools')
@@ -70,12 +76,24 @@ def test_typing_gate_blocks_root_change_when_base_root_present(tmp_path: Path) -
     assert len(_root_failures(failures)) == 1
 
 
-def test_fail_loud_gate_allows_root_rename_when_base_root_absent() -> None:
+def test_typing_gate_blocks_root_rename_with_total_change(tmp_path: Path) -> None:
+    typing_gate = _load('typing_gate')
+    base_path = tmp_path / 'base_budget.json'
+    base_path.write_text(json.dumps(_budget(TYPING_BUDGET, _absent_root(tmp_path))))
+    head_budget = _budget(TYPING_BUDGET, 'tools')
+    head_budget['pyright_errors'] = {'total': 4}
+
+    failures = typing_gate.gate_budget_source(str(base_path), False, head_budget)
+
+    assert any('totals-neutral' in failure for failure in _root_failures(failures))
+
+
+def test_fail_loud_gate_allows_root_rename_when_base_root_absent(tmp_path: Path) -> None:
     fail_loud_gate = _load('fail_loud_gate')
 
     failures = fail_loud_gate.gate(
         _budget(FAIL_LOUD_BUDGET, 'tools'),
-        _budget(FAIL_LOUD_BUDGET, 'renamed_away_pkg_xyz'),
+        _budget(FAIL_LOUD_BUDGET, _absent_root(tmp_path)),
     )
 
     assert _root_failures(failures) == []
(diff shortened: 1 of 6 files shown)


tests/tools/test_budget_source_ratchet.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_typing_gate_allows_root_rename_when_base_root_absent(tmp_path: Path) -> None:
+    typing_gate = _load('typing_gate')
+    base_path = tmp_path / 'base_budget.json'
+    base_path.write_text(json.dumps(_budget(TYPING_BUDGET, 'renamed_away_pkg_xyz')))
+
+    failures = typing_gate.gate_budget_source(
+        str(base_path), False, _budget(TYPING_BUDGET, 'tools')
+    )
```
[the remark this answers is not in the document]
zero-bang: +1 — agree this is worth tightening. The failure would be silent and confusing (a genuinely correct gate flags red only because someone unrelated added a dir with that exact name). Deriving the absent base_root from tmp_path (which is guaranteed unique) and asserting `not (REPO_ROOT / base_root).is_dir()` before writing the budget makes the intent self-verifying. Minor, non-blocking.

tools/typing_gate.py
```diff
@@ -632,11 +632,15 @@ def gate_budget_source(
     #   Adding an exclude hides files from the ratchet.
     base_root = base_budget.get('package_root')
     head_root = head_budget.get('package_root')
```
zero-bang, line 653: The allowance is gated only on the *old* root path being absent from the head tree — it never checks that the new root covers the same files. That leaves a narrowing channel that didn't exist before this PR:

1. `git mv tdw_control_plane tdw_control_plane_new` (all violating files move, intact)
2. set `package_root` to `tdw_control_plane_new/clean_subdir` (a subtree with 0 violations) and lower the totals

`tdw_control_plane` is now absent → `(REPO_ROOT / base_root).is_dir()` is False → the root change is **allowed**. The gate then scans only `clean_subdir`; the violating files under `tdw_control_plane_new/` survive in the tree but escape the ratchet. I checked the rest of `gate_budget_source` and `gate_files_analyzed` — nothing compares against the base file set, so nothing catches the orphaned files.

Contrast with the deletion case the PR reasons about: deleting the old root removes the violating files from the repo entirely, so they can't hide. A rename keeps them. So this *is* a new evasion channel, and 'the scan surface cannot be narrowed by the PR it gates' is no longer strictly true.

The gate can't verify a pure rename from the budget JSON alone (it has no view of the base tree's file list). The clean, verifiable fix is to keep the root pinned in any gating PR and require the rename to land in a separate PR that changes no totals — i.e. don't allow root change and total decreases in the same PR. If you'd rather accept the residual risk here, please narrow the claim in the description/CHANGELOG accordingly rather than stating no new channel exists.

tools/fail_loud_gate.py
```diff
@@ -314,11 +314,16 @@ def gate(
     # Structural: head must preserve package_root, must not add excludes,
     # must preserve every category key, must not raise any total.
     if budget_base is not None:
-        if budget_head.get('package_root') != budget_base.get('package_root'):
+        base_root = budget_base.get('package_root')
+        if budget_head.get('package_root') != base_root and (
```
zero-bang: Same narrowing channel as flagged in `typing_gate.py:634` — the allowance keys only on `(REPO_ROOT / base_root).is_dir()`, so a `git mv` of the package to a new parent plus a `package_root` pointed at a clean subtree of it is accepted while violating siblings remain in the tree outside the scan surface (totals-only-fall and the excludes-subset check don't catch orphaned files). Same fix applies: split the rename into its own non-gating PR, or narrow the 'no new evasion channel' claim.

zero-bang (changes requested):
Reviewed the full diff, both gate implementations end-to-end, and the actual checked-in budgets (`package_root: tdw_control_plane`). The mechanism, tests, CHANGELOG, version bump (2.0.1→2.1.0, correct minor for a new allowance), and the CI wiring (new test file added to `pr_checks_ruleset`) are all sound, and all 10 checks pass.

One blocking concern: the relaxation is scoped by *old-root-absence* alone, which permits narrowing via a subtree of a renamed directory. This reintroduces the exact scan-surface narrowing the ratchet exists to prevent, so the PR's central claim that it 'adds no new evasion channel' does not hold as written. Detail inline on both gates. Also +1'd Copilot's (non-blocking) test-brittleness thread.

Forward-looking: the next predictable failure mode is a legitimate refactor that *splits* `tdw_control_plane/` into a renamed parent with the package under a subdir — it would pass the gate while silently dropping the siblings from coverage, and no one would notice until errors accumulate unguarded.

tools/typing_gate.py
mikkokotila: Confirmed — the mv-to-new-parent + clean-subtree attack was real and my 'no new evasion channel' claim was wrong. Fixed in 4992c05 per your prescription: a root change is now legal only when the base root is absent AND `_ratchet_totals(base) == _ratchet_totals(head)` (every `patterns.*.total`, `pyright_errors.total`, `any_references.total` identical) — no ratchet movement can ride a rename; totals move in separate PRs. New test `test_typing_gate_blocks_root_rename_with_total_change` covers it. The residual (relocating files outside the scan surface, which any PR could always do by moving files out of the root) is now stated explicitly in the CHANGELOG with operator review named as the backstop.

tools/fail_loud_gate.py
mikkokotila: Fixed in 4992c05, same as typing_gate: `_category_totals(base) == _category_totals(head)` required for any root change, tested by `test_fail_loud_gate_blocks_root_rename_with_total_change`; claim narrowed in CHANGELOG.

tests/tools/test_budget_source_ratchet.py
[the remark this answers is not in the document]
mikkokotila: Fixed in 4992c05: `_absent_root(tmp_path)` derives the base root from tmp_path.name (unique per run) and asserts `not (REPO_ROOT / root).is_dir()` before use — self-verifying, no hardcoded name.

tests/tools/test_budget_source_ratchet.py
```diff
@@ -0,0 +1,92 @@
...
+    assert len(_root_failures(failures)) == 1
+
+
+def test_fail_loud_gate_allows_root_rename_when_base_root_absent() -> None:
+    fail_loud_gate = _load('fail_loud_gate')
+
+    failures = fail_loud_gate.gate(
+        _budget(FAIL_LOUD_BUDGET, 'tools'),
+        _budget(FAIL_LOUD_BUDGET, 'renamed_away_pkg_xyz'),
+    )
```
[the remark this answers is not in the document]
mikkokotila: Fixed in 4992c05 via the same `_absent_root(tmp_path)` helper (unique per run, asserted absent).

zero-bang (approved):

Issue #275, opened 2026-07-06
tracker: no-false-green, single-truth, self-heal (a/b/c program)


mikkokotila:
Tracker for the truth-and-self-heal program. Target properties, per operator:

- **(a) No false green** — the system must never look healthy when data has a gap, a run was lost, a tick was missed, or a delivery failed. *Absence of a red signal counts as false green.*
- **(b) Single truth** — Dagster/dagit is THE authoritative record: completeness, freshness, and delivery legible in dagit itself, not in logs, ad-hoc SQL, or external systems.
- **(c) Self-heal** — any gap is detected and repaired by the system itself, with a lookback horizon longer than any plausible outage.

Findings from the 2026-07-06 adversarial gap analysis (5 agents, 318 evidence calls against code + live production; evidence citations in each item). Items become slices under this tracker; tick when the closing slice merges.

## 🚨 P0 — security emergency (fix independent of the program)

- [x] **E1. Dagit publicly writable from the internet.** — ✅ DONE 2026-07-08: bound to `127.0.0.1` via PR #277 (deploy verified: external refused, tunnel access works, code location loads). Recent-run foreign-activity scan clean (0 manual launches / last 2000 runs). Original marker: — 🔒 live exposure CLOSED 2026-07-06 (dagit container stopped; pipeline unaffected). Durable fix in PR #277 (slice #276): bind `127.0.0.1:4000` + tunnel + a CI guard against any non-loopback deploy-compose port. Blocked on #279 (below); merges after it. Run-history audit for foreign activity still TODO.
- [x] **E1b. Law-9 ruleset gate blocked repo-wide** (found while landing E1) — GitHub started returning `dismissal_restriction` on live ruleset 5406599 after #274, drifting from the checked-in snapshot; every open PR failed `pr_checks_ruleset`. Fixed in PR #279 (slice #278): snapshot + 4 gate fixtures synced (disabled/default value, no protection change). Merge FIRST to unblock #277.
[...]

PR #277, opened 2026-07-08
fix(deploy): bind dagit to loopback so it is not publicly exposed

mikkokotila:

## What does this PR change?

E1 of tracker #275 (security P0). The deploy dagit webserver was published on `0.0.0.0:4000` since 2026-04-21 — no auth, externally verified reachable with full GraphQL mutation access (launch/terminate/delete runs, wipe assets, stop schedules). This binds it to `127.0.0.1:4000:3000`, matching how ClickHouse is already bound in the same file; operator access moves to an SSH tunnel (`ssh -L 4000:localhost:4000 root@37.27.112.167`, or `LocalForward 4000 localhost:4000` in `~/.ssh/config`). A new `tests/tools/test_deploy_compose_bind.py` (added to `pr_checks_ruleset`) fails any deploy-compose published port not bound to `127.0.0.1`, so nothing can silently re-expose a service. Version 3.0.1 → 3.0.2.

The live exposure was already closed out-of-band ahead of this PR (dagit container stopped — pipeline unaffected, daemon independent); this PR makes the fix durable, and the deploy restarts dagit loopback-bound.

## Checklist

- [x] I ran the project's test suite locally without errors (where applicable) — `pytest tests/tools -q`: 36 passed; `ruff check tools tests/tools`: clean

Closes #276

### DIFF
--- a/.github/workflows/pr_checks_ruleset.yml
+++ b/.github/workflows/pr_checks_ruleset.yml
@@ -29,7 +29,7 @@ jobs:
         run: .venv-ruleset/bin/python -m pip install --upgrade pip pytest 'ruff==0.15.11'
 
       - name: Run CI contract tests
-        run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py tests/tools/test_privileged_ruleset_audit.py tests/tools/test_budget_source_ratchet.py -q
+        run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py tests/tools/test_privileged_ruleset_audit.py tests/tools/test_budget_source_ratchet.py tests/tools/test_deploy_compose_bind.py -q
 
       - name: Compare live ruleset to checked-in snapshot
         env:
--- a/.github/workflows/pr_checks_ruleset.yml
+++ b/.github/workflows/pr_checks_ruleset.yml
@@ -26,7 +26,7 @@ jobs:
         run: python -m venv .venv-ruleset
 
       - name: Install contract dependencies
-        run: .venv-ruleset/bin/python -m pip install --upgrade pip pytest pyyaml 'ruff==0.15.11'
+        run: .venv-ruleset/bin/python -m pip install --upgrade pip pytest 'ruff==0.15.11'
 
       - name: Run CI contract tests
         run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py tests/tools/test_privileged_ruleset_audit.py tests/tools/test_budget_source_ratchet.py tests/tools/test_deploy_compose_bind.py -q
--- a/.github/workflows/pr_checks_ruleset.yml
+++ b/.github/workflows/pr_checks_ruleset.yml
@@ -26,7 +26,7 @@ jobs:
         run: python -m venv .venv-ruleset
 
       - name: Install contract dependencies
-        run: .venv-ruleset/bin/python -m pip install --upgrade pip pytest 'ruff==0.15.11'
+        run: .venv-ruleset/bin/python -m pip install --upgrade pip pytest pyyaml 'ruff==0.15.11'
 
       - name: Run CI contract tests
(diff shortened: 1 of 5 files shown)


.github/workflows/pr_checks_ruleset.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         run: .venv-ruleset/bin/python -m pip install --upgrade pip pytest 'ruff==0.15.11'
 
       - name: Run CI contract tests
-        run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py tests/tools/test_privileged_ruleset_audit.py tests/tools/test_budget_source_ratchet.py -q
+        run: .venv-ruleset/bin/python -m pytest tests/tools/test_cc_gate.py tests/tools/test_ruleset_gate.py tests/tools/test_ci_contract.py tests/tools/test_lint_ci_contract.py tests/tools/test_tests_ci_contract.py tests/tools/test_agents_contract.py tests/tools/test_privileged_ruleset_audit.py tests/tools/test_budget_source_ratchet.py tests/tools/test_deploy_compose_bind.py -q
```
zero-bang, line 32: Blocking — this is why `pr_checks_ruleset` is red. This job's venv is installed with only `pip pytest 'ruff==0.15.11'` (line 29); it has no PyYAML. The new `test_deploy_compose_bind.py` does `import yaml`, so pytest fails at collection (`ModuleNotFoundError: No module named 'yaml'`, exit 2) and the whole contract suite never runs.

Two fixes:
- **Preferred (keeps the contract venv minimal, matching the other pure-stdlib gates):** drop the PyYAML dependency in the test and parse the port lines yourself. The compose port block is simple short-form strings, so a tiny regex/line scan over `docker-compose.deploy.yml` is enough — no YAML parser needed.
- **Alternative:** add `pyyaml` to the install step on line 29 (`... pytest pyyaml 'ruff==0.15.11'`). Simpler, but pulls a third-party dep into a venv that was deliberately stdlib-only.

tests/tools/test_deploy_compose_bind.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Final
+
+import yaml
```
zero-bang: This `import yaml` is the direct cause of the required-check failure: the `pr_checks_ruleset` venv doesn't install PyYAML (see the workflow comment). Prefer parsing the compose file's short-form port strings with the stdlib to keep this gate dependency-free like the other contract tests, or add `pyyaml` to the ruleset job's install step.

zero-bang (changes requested):
The security fix is correct and well-scoped: `docker-compose.deploy.yml` now binds dagit as `127.0.0.1:4000:3000`, matching how ClickHouse is already bound, and the new guard test is a good durable regression net. CHANGELOG and version bump are in order.

However, **the required `pr_checks_ruleset` check is failing (blocking)**, so this can't merge as-is. The new test is wired into the `pr_checks_ruleset` job, whose venv installs only `pytest` + `ruff` — no PyYAML — so `import yaml` raises `ModuleNotFoundError` and pytest aborts collection with exit code 2:

```
ERROR collecting tests/tools/test_deploy_compose_bind.py
E   ModuleNotFoundError: No module named 'yaml'
```

Every other contract test in that job is pure-stdlib by design, and the full `.[dev]` env (`pr_checks_tests`) only runs `tests/origo_source_native`, so the minimal venv is the test's only CI home. Fix inline below. Once the check is green this is an approve.

Status ledger: `pr_checks_ruleset` = fail (root cause identified above, blocking); CodeQL / `pr_checks_tests` / `pr_checks_typing` still pending; `pr_checks_cc/lint/version/slice/fail_loud` = pass. No other open review threads.

Forward-looking: the guard only inspects short-form string port mappings (`spec.get('ports', [])` as `str`). If someone later adds a long-form port entry (`{published: 4000, host_ip: 0.0.0.0}`), `str(mapping)` won't start with `127.0.0.1:` — so it would actually still be caught here, but a long-form entry *without* a published port (target-only) would be silently skipped. Worth a comment noting the test assumes short-form syntax.

.github/workflows/pr_checks_ruleset.yml
mikkokotila: Fixed in 4a2cb04 (took your preferred option): reverted the pyyaml install on line 29 — the ruleset venv is stdlib-only again. The test now hand-parses the compose ports, so nothing imports yaml and the contract suite collects cleanly.

tests/tools/test_deploy_compose_bind.py
mikkokotila: Fixed in 4a2cb04: no more `import yaml`. The port block is parsed with a small stdlib walker (services -> service -> ports indentation), matching the other pure-stdlib gates.

tests/tools/test_deploy_compose_bind.py
```diff
@@ -0,0 +1,38 @@
...
+        f'{service}: {mapping}'
+        for service, mapping in _published_ports()
+        if not mapping.startswith(f'{LOOPBACK}:')
+    ]
+    assert exposed == []
+
+
+def test_dagit_bound_to_loopback() -> None:
+    dagit_ports = [mapping for service, mapping in _published_ports() if service == 'dagit']
+    assert dagit_ports == [f'{LOOPBACK}:4000:3000']
```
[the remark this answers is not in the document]
mikkokotila: Fixed in 4a2cb04: the guard now normalises short form (`ip:host:container`), block long form (`{target, published, host_ip}` across lines), and inline flow maps, and strips `/tcp`-style protocol suffixes. Verified against 6 adversarial fixtures: loopback short form passes; a public `4000:3000` and a long-form entry missing `host_ip` are both caught; protocol-suffixed and inline-map loopback entries pass.

zero-bang (approved):

PR #281, opened 2026-07-10
feat(origo): add hourly gap-repair schedules for the spot and futures daily pipelines

mikkokotila:

## What does this PR change?

Tracker #275 item 17 (slice #280) — the highest-value self-heal gap: the daily spot/futures pipelines had a lookback of zero (one fire-and-forget tick per day), so a killed run left a permanent hole (the 2026-07-03 futures incident needed a manual backfill). This adds `origo/utils/daily_gap_repair.py` (pure, unit-tested gap detection) and two hourly schedules (`30 * * * *`, RUNNING) targeting the existing partitioned jobs: a day is repairable iff its ingestion ledger has no `status='success'` row AND its Binance archive exists (same `.CHECKSUM` probe the assets use), within a 14-day window ending at **today−2** so the repair never races the regular 04:00/10:00 ticks over yesterday. Run keys embed the tick date (`daily_gap_repair:<market>:<day>:<today>`) — at most one attempt per day per gap, max 3 repair runs per tick. Re-runs are idempotent (the assets delete-then-insert their partition and rewrite the ledger row), so this is safe without the atomic-replacement work (§4), which remains a prerequisite only for run-level retries. Spot's default archive URL is promoted to a public constant (`DEFAULT_BINANCE_SPOT_DAILY_TRADES_BASE_URL`); futures already had one. Four new tests run against the real ClickHouse + fixture-archive harness. Version 3.0.3 → 3.1.0.

## Checklist

- [x] I ran the project's test suite locally without errors (where applicable) — `pytest tests/origo_source_native -q --maxfail=1`: 153 passed; `pytest tests/tools -q`: 36 passed; `ruff check tools tests/tools`: clean

Closes #280

### DIFF
--- a/origo/utils/daily_gap_repair.py
+++ b/origo/utils/daily_gap_repair.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Callable
+from collections.abc import Callable, Set as AbstractSet
 from dataclasses import dataclass
 from datetime import date, timedelta
 
@@ -32,7 +32,10 @@ def archive_available(base_url: str, day: date) -> bool:
         response = requests.get(checksum_url, timeout=ARCHIVE_PROBE_TIMEOUT_SECONDS)
     except requests.RequestException:
         return False
-    return response.status_code == 200
+    try:
+        return response.status_code == 200
+    finally:
+        response.close()
 
 
 def repair_window(today: date, earliest_partition: date) -> tuple[date, date] | None:
@@ -78,7 +81,16 @@ def repairable_gap_days(
     database: str,
     spec: DailyGapRepairSpec,
     today: date,
+    active_partition_days: AbstractSet[date],
 ) -> list[date]:
+    """Ledger-absent days with an available archive and no in-flight run.
+
+    ``active_partition_days`` are partitions with an in-progress run of the
+    same job — including a regular daily tick still inside its op-retry
+    backoff (RetryPolicy holds the run in STARTED for up to ~23h) — which
+    must never be raced by a concurrent repair run of the non-atomic
+    delete-then-insert assets.
+    """
     window = repair_window(today, spec.earliest_partition)
     if window is None:
         return []
@@ -88,7 +100,7 @@ def repairable_gap_days(
     gaps: list[date] = []
     day = start
     while day <= end and len(gaps) < MAX_GAP_REPAIR_RUNS_PER_TICK:
-        if day not in loaded and archive_available(base_url, day):
+        if day not in loaded and day not in active_partition_days and archive_available(base_url, day):
             gaps.append(day)
         day += timedelta(days=1)
     return gaps
@@ -99,6 +111,7 @@ def gap_repair_run_requests(
     database: str,
     spec: DailyGapRepairSpec,
     today: date,
+    active_partition_days: AbstractSet[date],
 ) -> list[RunRequest] | SkipReason:
     """One RunRequest per repairable gap day, keyed once per day per gap.
 
@@ -106,7 +119,7 @@ def gap_repair_run_requests(
     at most once per day until its ledger row appears, without ever
     re-running a partition the same day it was already requested.
     """
-    gaps = repairable_gap_days(client, database, spec, today)
+    gaps = repairable_gap_days(client, database, spec, today, active_partition_days)
     if not gaps:
         return SkipReason(f'{spec.market}: no repairable daily gaps in lookback.')
     return [
--- a/origo/utils/daily_gap_repair.py
+++ b/origo/utils/daily_gap_repair.py
@@ -1,0 +1,118 @@
+from __future__ import annotations
+
+from collections.abc import Callable
+from dataclasses import dataclass
+from datetime import date, timedelta
+
+import requests
+from dagster import RunRequest, SkipReason
+
+from origo.assets.create_origo_database import ClickHouseClientProtocol
+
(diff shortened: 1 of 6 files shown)


origo/utils/daily_gap_repair.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def repair_window(today: date, earliest_partition: date) -> tuple[date, date] | None:
+    """[today - lookback, today - 2]; yesterday is left to the regular daily tick.
+
+    Ending at today - 2 means the hourly repair never races the 04:00/10:00
+    schedules over yesterday's partition: a missed yesterday becomes
+    repairable the following day.
+    """
+    end = today - timedelta(days=2)
```
zero-bang, line 48: Concurrency race against a still-retrying regular run. The docstring's "never races the daily ticks" guarantee holds for the *scheduled* tick, but not for its **retry lifetime**. The daily assets carry `RetryPolicy(max_retries=23, delay=3600)` (~23h horizon). A run for partition `D-1` launched by the regular tick on day `D` (04:00 spot / 10:00 futures) that keeps failing stays in retry backoff well into day `D+1`. On `D+1`, `end = today - 2 = D-1`, so `D-1` is now inside the repair window; it still has no `status='success'` ledger row (still failing) and its archive exists, so gap repair launches a **second** run for `D-1` at 00:30 `D+1` — before retries are exhausted (~03:00/09:00+ `D+1`).

The gap-repair run key (`daily_gap_repair:...`) is a different namespace from the regular tick's key, so Dagster does not dedup them, and `dagster.yaml` configures no run coordinator / tag concurrency, so both runs execute simultaneously. `_process_day` is not atomic (count → `DELETE` → `INSERT` → verify `inserted_count == len(rows)`): two concurrent runs can interleave into duplicated partition rows plus a `success` ledger row whose `inserted_row_count` disagrees with the table — silent corruption that also feeds the downstream kline/dollar-bar assets. That is worse than the loud gap it repairs.

This is exactly the run-level-retry case the PR defers to §4 (atomic replacement); the `today-2` boundary was assumed to prevent it, but the retry horizon crosses the day boundary. Concrete fixes, either: (a) push the window end past the retry horizon — `end = today - timedelta(days=3)` — so a still-retrying partition has exhausted its ~23h of retries before becoming repair-eligible; or (b) add a per-partition run-concurrency tag / skip days that have an in-progress or up-for-retry run. (Worth verifying Dagster keeps the run in `STARTED` during retry backoff, but that is the standard step-retry behavior.)

zero-bang (changes requested):
Solid, well-scoped self-heal. The pure gap-detection module is cleanly separated and unit-tested; keying by `today` for once-per-day dedup, gating on the `.CHECKSUM` probe, and the `today-2` window are all thoughtful, and running the full partitioned job (so downstream klines re-materialize too) is the right call.

One blocking concern: the `today-2` window boundary does not actually clear the regular pipeline's *retry horizon*, opening a concurrent-run race against the non-atomic `_process_day` (inline). Because there is no run coordinator / per-partition concurrency limit configured (`dagster.yaml` uses the default `DefaultRunCoordinator`), two runs for the same partition can execute at once — which the PR's own scope note assumes cannot happen.

Open-threads ledger:
- No prior reviewer comments to reconcile.
- CI: `pr_checks_tests`, `pr_checks_typing`, and CodeQL are still *pending* at review time; lint/version/slice/cc/ruleset/fail_loud are green. Position: these must be green before merge, but I have no reason to believe they'll fail (author reports local pass); not counting them as a separate finding, just flagging they are not yet complete.

Forward-looking: the next predictable failure mode is a persistently-failing daily tick (e.g. repeated checksum mismatch or ClickHouse unavailable for hours) rather than a cleanly killed run — that is the case that trips the race below.

origo/utils/daily_gap_repair.py
mikkokotila: Confirmed and fixed in 68bfda8, taking your option (b): `repairable_gap_days`/`gap_repair_run_requests` now take `active_partition_days`, and the schedules feed them from a new `_active_partition_days(context, job_name)` that queries the instance for in-progress runs (QUEUED/NOT_STARTED/STARTING/STARTED/CANCELING) of the same job and excludes their partitions — so a regular run held in STARTED by its 23h op-retry backoff can never be raced, regardless of retry-policy changes. (Chose (b) over (a) because a static today−3 boundary would silently re-open if anyone raises max_retries past the boundary, and it costs a day of heal latency.) The docstrings and slice #280 now describe the two-layer exclusion honestly: window boundary for the scheduled tick, in-flight exclusion for the retry lifetime. New test `test_day_with_in_progress_run_is_not_repaired` covers it; suite 154 green.

origo/utils/daily_gap_repair.py
```diff
@@ -0,0 +1,118 @@
...
+    return f'BTCUSDT-trades-{day.isoformat()}.zip'
+
+
+def archive_available(base_url: str, day: date) -> bool:
+    checksum_url = f'{base_url}{source_filename(day)}.CHECKSUM'
+    try:
+        response = requests.get(checksum_url, timeout=ARCHIVE_PROBE_TIMEOUT_SECONDS)
+    except requests.RequestException:
+        return False
+    return response.status_code == 200
```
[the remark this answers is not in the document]
mikkokotila: Fixed in 68bfda8: the `.CHECKSUM` probe response is now closed in a `finally` block, mirroring `_depth_source_has_rows`.

zero-bang (approved):

PR #283, opened 2026-07-10
feat(origo): enable run monitoring so orphaned runs fail loudly

mikkokotila:

## What does this PR change?

Tracker #275 item 1 (slice #282): enables Dagster run monitoring in `dagster.yaml`. Two zombie classes we actually hit become loud failures: runs stuck in `STARTING` fail after 300s (the bar-store zombies sat 19 days), and any run exceeding `max_runtime_seconds: 93600` (26h) is terminated (the 07-03 futures run sat `STARTED` for 6h until manually cancelled). 26h is deliberately above the daily assets' 23h op-retry envelope — `RetryPolicy(max_retries=23, delay=3600)` — so runs retrying as designed are never reaped; per-run override via the canonical `dagster/max_runtime` tag. Verified against the installed dagster 1.13.12: `DefaultRunLauncher` has no worker health checks, so start-timeout and max-runtime are the effective mechanisms (both confirmed present in the monitoring daemon source), and the config block validates against `dagster_instance_config`.

Three contract tests: monitoring enabled with guards; **max runtime clears the in-code retry envelope** (the test reads the actual `RetryPolicy` off the asset, so the contract cuts both ways — raising retries past the runtime cap also fails CI); and `run_retries` stays absent until partition replacement is atomic (#275 item 21's sequencing, now mechanical). Version 3.1.0 → 3.2.0.

Honest capability bound (per review): with `DefaultRunLauncher` (no worker health checks), STARTING orphans fail at 300s, but a dead STARTED worker turns red only at the 26h bound — a 07-03-class orphan is bounded (~26h + 1h grace + next repair tick ≈ ≤27.5h total heal), not instantly detected; real orphan detection needs a health-checking launcher (tracker follow-up). Gap repair additionally excludes partitions with a terminal run in the last hour (`REPAIR_TERMINATION_GRACE_SECONDS`), because a timed-out run is force-marked FAILED without confirmed worker exit.

## Checklist

- [x] I ran the project's test suite locally without errors (where applicable) — `pytest tests/origo_source_native -q --maxfail=1`: 159 passed; `ruff check tools tests/tools`: clean; `dagster_instance_config` validates the yaml

Closes #282


tests/origo_source_native/test_dagster_instance_config.py
```diff
@@ -0,0 +1,57 @@
...
+    config = _instance_config()
+    run_monitoring = config['run_monitoring']
+    assert isinstance(run_monitoring, dict)
+    max_runtime = run_monitoring['max_runtime_seconds']
+    assert isinstance(max_runtime, int)
+
+    retry_policy = insert_daily_binance_spot_trades_to_origo.op.retry_policy
+    assert retry_policy is not None
+    assert retry_policy.delay is not None
+    retry_envelope_seconds = int(retry_policy.max_retries * retry_policy.delay)
```
zero-bang: The envelope this compares against is retry *delays* only — `max_retries * delay` = 23 * 3600 = 82800s (23h). It omits the execution time of the 24 attempts themselves. Since op retries run in-process (the run stays STARTED across all attempts), the real wall-clock a legit run can consume is 82800s + sum of per-attempt execution. With only ~3h (10800s) of headroom to 93600s, an average attempt cost above ~7.5 min across 24 attempts pushes a fully-retrying run past the cap — and it gets reaped while behaving exactly as designed, yet this test stays green. So the docstring's "must never be reaped" guarantee is stronger than what's asserted. Suggest either folding a realistic per-attempt execution allowance into the compared value, or softening the docstring to state that the 3h margin is the shared execution budget across all 24 attempts. Separately, this only reads the spot asset; daily_futures_trades_to_origo.py:299 carries an identical policy but isn't covered — parametrizing over both would keep the contract honest if the two ever diverge.

dagster.yaml
```diff
@@ -21,3 +21,10 @@ schedule_storage:
   class: SqliteScheduleStorage
   config:
     base_dir: /opt/dagster-instance
+
+run_monitoring:
+  enabled: true
+  start_timeout_seconds: 300
+  cancel_timeout_seconds: 180
+  max_runtime_seconds: 93600
+  poll_interval_seconds: 120
```
zero-bang, line 30: This cap is instance-wide, so it also bounds assets with no retry policy — notably build_bar_store_arrow and build_depth_snapshot_store_arrow, which do full historical (re)builds. Worth confirming a from-scratch rebuild of those stores never legitimately exceeds 26h; if it can, run monitoring will terminate it mid-build. If that's a real risk, a per-run `dagster/max_runtime_seconds` tag on those assets is the escape hatch the PR description already references.

zero-bang (commented):
Reviewed the full diff. The config block is well-formed and every key (enabled, start_timeout_seconds, cancel_timeout_seconds, max_runtime_seconds, poll_interval_seconds) is a valid `run_monitoring` field; poll (120s) < start_timeout (300s) so a STARTING zombie is caught promptly. The three contract tests are meaningful — the `.op.retry_policy` access matches the established pattern in test_daily_gap_repair.py and the futures template — and CHANGELOG/version follow the repo's latest-on-top convention. Checks: cc, fail_loud, slice, lint, version pass; tests/typing/CodeQL/ruleset still pending (not red — not blocking).

Not requesting changes. Two non-blocking observations below, plus a forward-looking question: the next predictable failure mode isn't a stuck run, it's a legitimately-slow one — a run that exhausts many retries with non-trivial per-attempt work, or a large store rebuild, hitting the instance-wide 26h cap while behaving as designed.

dagster.yaml
```diff
@@ -21,3 +21,10 @@ schedule_storage:
   class: SqliteScheduleStorage
   config:
     base_dir: /opt/dagster-instance
+
+run_monitoring:
+  enabled: true
+  start_timeout_seconds: 300
+  cancel_timeout_seconds: 180
+  max_runtime_seconds: 93600
```
mikkokotila, line 29: [P1] STARTED deploy-orphans still wait up to 26h

Production runs Dagster 1.13.13 with `DefaultRunLauncher.supports_check_run_worker_health == False`. `monitor_started_run` therefore only applies `check_run_timeout`; the 120s poll does not detect a dead STARTED worker before total run age exceeds 93,600s. The 2026-07-03 incident was canceled at 22,782s, so this config would have left it STARTED—and excluded by `_active_partition_days` from hourly repair—for roughly another 19h40m. This contradicts the PR body's claim that a deploy-orphaned run turns red within the poll interval. Either narrow the capability to a 26h bound or add worker-health detection that distinguishes an orphan from legitimate retries.

mikkokotila, line 29: [P1] Keep repair blocked until the timed-out worker exits

Dagster 1.13.13 `check_run_timeout` sends `terminate`, then immediately reports/force-marks the run FAILED. `DefaultRunLauncher`'s successful `CancelExecution` only sets a termination event; it does not confirm subprocess exit, and a failed termination is force-marked too. `cancel_timeout_seconds` is bypassed on this path. Because `_active_partition_days` excludes only in-progress statuses, the hourly repair can launch the same non-atomic delete/insert partition while the original worker is still running or cleaning up. That recreates the #275 item 21 race this PR says it avoids. Add a proven worker-exit barrier or keep timed-out partitions excluded through a cancellation grace period.

tests/origo_source_native/test_dagster_instance_config.py
```diff
@@ -0,0 +1,57 @@
...
+import yaml
+
+from origo.assets.daily_trades_to_origo import insert_daily_binance_spot_trades_to_origo
+
+REPO_ROOT: Final[Path] = Path(__file__).resolve().parents[2]
+DAGSTER_YAML: Final[Path] = REPO_ROOT / 'dagster.yaml'
+
+
+def _instance_config() -> dict[str, object]:
+    loaded = yaml.safe_load(DAGSTER_YAML.read_text(encoding='utf-8'))
```
mikkokotila: [P2] Exercise Dagster's instance-config schema

`yaml.safe_load` proves only YAML syntax. A misspelled or unknown `run_monitoring` field can keep all three new tests green while `dagster-daemon` rejects `dagster.yaml` at startup. The PR says `dagster_instance_config` was validated, but the regression gate never calls it. Load this file through Dagster's config loader and assert against the validated mapping.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v3.2.0 on July 10, 2026
+- Enable Dagster run monitoring: runs stuck in STARTING fail after 300s, and any run exceeding 26h (above the daily assets' 23h op-retry envelope; per-run override via the dagster/max_runtime_seconds tag) is terminated — orphaned runs now turn red instead of sitting in-progress forever (tracker #275 item 1). Run retries stay disabled until partition replacement is atomic (item 21).
```
mikkokotila: [P2] Name the canonical per-run override

Dagster 1.13.13 defines `MAX_RUNTIME_SECONDS_TAG` as `dagster/max_runtime`. `dagster/max_runtime_seconds` works only because `check_run_timeout` currently carries an explicit compatibility fallback; it is not the canonical tag the changelog calls documented. Point operators to `dagster/max_runtime` so the advertised escape hatch does not depend on an internal alias.

mikkokotila (commented):
Four actionable findings; details inline.

tests/origo_source_native/test_dagster_instance_config.py
mikkokotila: Fixed in c1abfd6: the compared value is now `max_retries * delay + (max_retries + 1) * PER_ATTEMPT_EXECUTION_BUDGET_SECONDS` with the budget explicit at 450s — 26h decomposes exactly as 23h delays + 24×450s execution, so the docstring's guarantee and the assertion now say the same thing (a fully-retrying run whose average attempt exceeds 450s is reaped by design, stated in the constant's comment). Also parametrized over both spot and futures assets per your note.

dagster.yaml
mikkokotila: Checked against production behavior: `build_bar_store_arrow` rebuilds per-series from the parquet mirror in seconds (48 runs observed post-deploy, all sub-minute) and `build_depth_snapshot_store_arrow` writes one minute-chunk per run — full-history rebuilds fan out per-partition rather than one long run, so nothing legitimately approaches 26h today. The canonical `dagster/max_runtime` tag (name fixed per the other thread) is the escape hatch if that ever changes.

tests/origo_source_native/test_dagster_instance_config.py
```diff
@@ -0,0 +1,57 @@
...
+
+def test_run_monitoring_is_enabled_with_zombie_guards() -> None:
+    config = _instance_config()
+    run_monitoring = config['run_monitoring']
+    assert isinstance(run_monitoring, dict)
+    assert run_monitoring['enabled'] is True
+    assert isinstance(run_monitoring['start_timeout_seconds'], int)
+    assert run_monitoring['start_timeout_seconds'] >= 60
+    assert isinstance(run_monitoring['max_runtime_seconds'], int)
+
```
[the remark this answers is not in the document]
mikkokotila: Fixed in c1abfd6: `test_run_monitoring_matches_zombie_guard_contract` now compares the whole `run_monitoring` mapping byte-exactly against `EXPECTED_RUN_MONITORING` (300/180/93600/120 + enabled), loaded through Dagster's own validator.

dagster.yaml
mikkokotila: Correct — the 'red within the poll interval' claim was wrong for STARTED orphans and is retracted. Fixed in c1abfd6 by narrowing the capability everywhere it was stated (PR body, CHANGELOG, slice #282): with DefaultRunLauncher (no `check_run_worker_health`), a dead STARTED worker turns red at the 26h bound, not sooner; STARTING orphans fail at 300s. A 07-03-class orphan is therefore bounded (~26h reap + 1h grace + next repair tick ≈ ≤27.5h total heal), not instantly detected. Real orphan detection needs a health-checking launcher — noted on tracker #275 as the follow-up beyond this slice's honest scope.

mikkokotila: Fixed in c1abfd6 with your grace-period option: `_active_partition_days` now returns in-progress partitions UNION partitions whose runs reached FAILURE/CANCELED within `REPAIR_TERMINATION_GRACE_SECONDS = 3600` (via `RunsFilter(updated_after=...)`), so a force-marked-FAILED run's partition stays repair-excluded for a full hour — at least one repair tick — while any lingering worker drains. Covered by `test_active_partition_days_covers_in_progress_and_recent_terminal` against a real ephemeral instance (STARTED excluded, fresh FAILURE excluded, SUCCESS not excluded).

tests/origo_source_native/test_dagster_instance_config.py
mikkokotila: Fixed in c1abfd6: all config tests now load `dagster.yaml` through `dagster_instance_config` (Dagster's schema validator) instead of `yaml.safe_load` — an unknown or misspelled `run_monitoring` field now fails the tests exactly as it would fail the daemon at startup.

CHANGELOG.md
mikkokotila: Fixed in c1abfd6: CHANGELOG, PR body, and slice now name the canonical `dagster/max_runtime` tag (the `_seconds` variant only works via an internal compatibility fallback, as you note).

zero-bang (approved):

PR #285, opened 2026-07-14
feat(origo): make daily partition writes atomic via a staging table

mikkokotila:

## What does this PR change?

Tracker #275 item 4 (slice #284): makes the daily spot/futures partition write atomic-in-practice. Today each asset does `DELETE day` then a **client-side INSERT of a full day's rows (millions, over minutes) straight into the live table** — the exact window in which a killed run (deploy, OOM, the 2026-07-03 incident) leaves a partial day with no ledger `success` row. New `origo/utils/atomic_day_write.replace_day_via_staging` builds the day in a per-day staging table (`<table>__staging_<yyyymmdd>`), **count-verifies it before touching the live table**, then promotes it with a synchronous day `DELETE` + an atomic `ALTER TABLE staging MOVE PARTITION ID '<yyyymm>' TO TABLE main` — a metadata part-move (not a splittable `INSERT..SELECT`, which ClickHouse divides at `max_insert_block_size` and can leave partially committed on cancellation, as review reproduced on the 24.3 image). MOVE appends the day into the live month partition, preserving the month's other days. The residual `DELETE`→`MOVE` gap leaves the day MISSING on a kill (never partial or duplicated), self-healed by daily gap-repair. The staging table is always dropped, and a stale one from a crashed prior attempt is dropped before rebuild.

Atomicity is in-practice, not engine-level (the monthly partitioning of these 9-year tables precludes day-granular `REPLACE PARTITION`; day partitioning would mean ~3,200 partitions — rejected). No partial/failed parse ever reaches the live table, and the promotion itself cannot commit a partial day. This is the prerequisite the run-monitoring slice (#283) deferred: item 21 (run-level retries) can follow once writes are safe to re-run.

Deterministic fake-client unit tests prove the live table is untouched on a staging-verify failure, the staging table is always dropped (success and error paths), and stale staging is dropped before rebuild; a real-ClickHouse test asserts no staging table leaks after a materialize; and the existing `test_reingest_same_day_preserves_exact_final_state` now exercises the staging swap end-to-end. Version 3.2.0 → 3.3.0.

## Checklist

- [x] I ran the project's test suite locally without errors (where applicable) — `pytest tests/origo_source_native -q --maxfail=1`: 165 passed; `ruff check tools tests/tools`: clean

Closes #284


origo/utils/atomic_day_write.py
```diff
@@ -0,0 +1,90 @@
...
+    before touching the live table, then swaps it in with a minimal
+    server-side DELETE + INSERT..SELECT. Returns the live row count for the
+    day, which the caller asserts against the source row count before writing
+    the ledger. The staging table is always dropped, and a stale staging
+    table from a crashed prior attempt is dropped before the rebuild.
+    """
+    staging = _staging_table_name(main_table, date_str)
+    column_list = ', '.join(columns)
+    try:
+        client.execute(f'DROP TABLE IF EXISTS {database}.{staging}')
```
zero-bang, line 69: Non-blocking / forward-looking: stale-staging cleanup only happens when *this exact day* is re-run (the leading `DROP IF EXISTS`) or when the `finally` runs. If a run is SIGKILL/OOM-killed mid-staging-INSERT **and that partition is never re-materialized**, its `<table>__staging_<yyyymmdd>` table leaks a partial day (potentially millions of rows) on disk indefinitely — nothing sweeps orphaned staging tables globally. In practice gap-repair/retries will re-run most days, so this is minor, but a periodic maintenance schedule that drops `%__staging_%` tables older than, say, a day would close the gap. Worth a tracker follow-up rather than a change here.

origo/utils/atomic_day_write.py
```diff
@@ -0,0 +1,90 @@
...
+        )
+        staged_count = _count_day(client, database, staging, date_str)
+        if staged_count != len(rows):
+            raise ValueError(
+                f'Staging row count mismatch for {date_str}: '
+                f'expected {len(rows)}, staged {staged_count}. Live table untouched.'
+            )
+
+        # Live table touched only from here: minimal, server-side, seconds.
+        client.execute(
```
zero-bang, line 85: Minor, re: the description's "sub-second server-side swap" claim: the `ALTER TABLE ... DELETE` is a mutation that rewrites the affected parts of the whole *monthly* partition, so for a dense recent month this is more likely tens of seconds than sub-second. This is not a regression (the old path issued the same `ALTER DELETE`), and the multi-minute client-side insert window is genuinely gone — just flagging that the residual window is dominated by this mutation, not the `INSERT..SELECT`, in case anyone later relies on the "seconds" figure for the run-retry timing in item 21.

zero-bang (approved):
Reviewed the full diff plus the surrounding asset/ledger flow and the two table DDLs. This is a clean, well-scoped change that does what it claims: the heavy client-side, multi-minute row insert now lands in a per-day `MergeTree` staging table (created via `CREATE TABLE ... AS`, which correctly inherits the schema/engine/partition/order/CODECs), is count-verified there, and only then is swapped into the live table via a server-side `DELETE + INSERT..SELECT`. The live table is genuinely untouched until the day is proven complete.

Things I verified and am satisfied with:
- Column constants (`SPOT_TRADE_COLUMNS` = 8 cols incl. `is_best_match`, `FUTURES_TRADE_COLUMNS` = 7 cols) exactly match the two table DDLs, so no column silently defaults.
- Both target tables are plain `MergeTree()` (not `Replicated`), so `CREATE TABLE ... AS` won't collide on a ZooKeeper path.
- The staged count check (`WHERE toDate(datetime)=...`) also catches any out-of-day source row before the live table is touched.
- Stale-staging handling: `finally` drops on both success and raised-error paths, and `DROP TABLE IF EXISTS` before rebuild covers a SIGKILL/OOM that skipped the `finally`. The op-level `RetryPolicy(max_retries=23)` composes safely with this — a retry after a crash in the residual window heals the day.
- Tests are meaningful: the fake-client tests prove the live table is untouched on a staging-verify failure and that staging is always dropped, and the real-CH test asserts no `__staging_` leak.

Residual window (DELETE done, INSERT..SELECT not yet) is the acknowledged in-practice tradeoff and is caught by the #283 monitoring + retries — I'm comfortable with it.

Two non-blocking notes inline. Approving.

Checks ledger: `pr_checks_lint`, `ruleset`, `version`, `slice`, `fail_loud`, `cc` are green; `tests`, `typing`, and CodeQL were still pending at review time — they should be green before merge, but none is red so none blocks the review.

origo/utils/atomic_day_write.py
```diff
@@ -0,0 +1,90 @@
...
+            f"""
+            ALTER TABLE {database}.{main_table}
+            DELETE WHERE toDate(datetime) = toDate('{date_str}')
+            """,
+            settings={'mutations_sync': 2},
+        )
+        client.execute(
+            f'INSERT INTO {database}.{main_table} ({column_list}) '
+            f'SELECT {column_list} FROM {database}.{staging}'
```
zero-bang: [P1] Promote the staged data with an atomic metadata operation

ClickHouse makes each inserted block atomic, not an arbitrary multi-block `INSERT … SELECT`. These daily datasets contain millions of rows, so this query is split at `max_insert_block_size` (~1,048,576). Against the project’s ClickHouse 24.3 image, I cancelled this exact promotion shape after its first block became visible; the live table retained 1,048,449 rows and one active part. The post-insert count never runs after a killed worker, so the PR still leaves the partial day it is intended to prevent—only for a shorter window. `ALTER TABLE <staging> MOVE PARTITION <month> TO TABLE <main>` after the synchronous day delete is a suitable metadata-level promotion here: I verified that a staging monthly partition containing only the replacement day appends to an existing destination month while preserving its other days. Please use an atomic partition/part promotion and add an interrupted-promotion regression test. ClickHouse documents the single-block limit and partition operations here: https://clickhouse.com/docs/knowledgebase/achieving-atomic-inserts

zero-bang (changes requested):
Blocking: the live promotion still commits per block and can leave a partial day after cancellation. All checks are green, but this defeats the central failure guarantee; see the inline reproduction.

origo/utils/atomic_day_write.py
mikkokotila: Confirmed and fixed in c2dd60e — your reproduction is exactly right; INSERT..SELECT is multi-block and my 'sub-second window' claim was wrong. The promotion is now a synchronous day `DELETE` followed by `ALTER TABLE {staging} MOVE PARTITION ID '<yyyymm>' TO TABLE {main}` (your suggested primitive). Verified against the real 24.3 image: `test_replacing_a_day_preserves_other_days_in_the_month` seeds 2024-01-01 and 2024-01-02 (same 202401 partition), re-ingests only the 2nd, and asserts day-1's rows are untouched and day-2 has no duplicate trade_ids — i.e. MOVE appends into the live month rather than replacing it. The interrupted-promotion case is now structural: MOVE is a single metadata part-move (asserted in `test_happy_path_promotes_via_atomic_move_partition` — no INSERT into the live table), so a kill can't commit a partial day; the only residual is a MISSING day in the DELETE→MOVE gap, which is never partial/duplicated and is healed by the daily gap-repair schedule (#280). CHANGELOG/PR/slice claims narrowed accordingly.

zero-bang (approved):

Issue #295, opened 2026-08-18
feat(origo): add a daily BTC briefing feed asset to origo/assets


mikkokotila:
### Slice ID

S295

### Slice Title

Add a daily BTC briefing feed asset to origo/assets

### Routing

> **Significance.** Routing is the minimum information needed to place this
> slice in the dependency graph. Parent and dependencies must be existing
> issue numbers; unlocks may be forward-looking. Prose description of the
> relationship is not accepted — only issue references. This section is not
> a planning document; it is a graph edge list.

Parent issue: none
Depends on:
- none
Unlocks:
- Vaquum/daily-briefing-btc-usd cutover from Coinbase public REST to origo

### Minimum Viable Capability

> **Significance.** A slice delivers exactly one capability. The capability
> must be expressible as a set of assertions that become mechanically true
> at merge. This is the answer to *what changes in the system that an
> external observer can verify by running a command?* Every line below
> must be evaluable as a shell command, SQL query, Python expression, or
> HTTP request that returns a definite result. One assertion per line.
>
> Accepted:
> - `clickhouse-client --query "SELECT countIf(stream_sequence IS NULL) FROM canonical_event_log"` returns `0`.
> - `python -c "import inspect, origo.events.writer as m; print(inspect.signature(m.CanonicalEventWriter.__init__))"` returns `(self, *, clickhouse_client)`.
> - `curl -sf https://origo.vaquum.fi/v1/health/schema | jq -r '.canonical_event_log_v1.columns | length'` returns `17`.
>
> Rejected:
> - *Freeze the current-truth read contract.*
> - *Strict CQRS separation.*
> - *Terminal proof obligations are complete.*

Assertions (one per line, each executable as-is):
- `python -c "import inspect, origo.assets.publish_btc_briefing_feed as m; print(inspect.signature(m.build_briefing_feed))"` returns `(client: _ClickHouseArrowClientProtocol, day: datetime.date) -> dict[str, object]`.
- `python -c "import origo.assets.publish_btc_briefing_feed as m; print(m.FEED_VERSION)"` returns `btc_briefing/1`.
- `python -c "import origo.assets.publish_btc_briefing_feed as m; print(sorted(m.FEED_SECTIONS))"` returns `['bars_15m', 'bars_1d', 'book_percentiles', 'book_series', 'book_sessions', 'volume_at_price']`.
- `pytest tests/origo_source_native/test_publish_btc_briefing_feed.py -q` returns exit code `0`.
[...]

PR #298, opened 2026-08-18
feat(origo): publish the daily BTC briefing feed to HuggingFace on the daily cadence

### LINKED ISSUE #297: feat(origo): publish the daily BTC briefing feed to HuggingFace on the daily cadence
### Slice ID

S297

### Slice Title

Publish the daily BTC briefing feed to HuggingFace on the daily cadence

### Routing

Parent issue: none
Depends on:
- #295
Unlocks:
- Vaquum/daily-briefing-btc-usd cutover from Coinbase public REST to origo (new-format first report, then deletion of the pre-cutover reports in that repository)

### Minimum Viable Capability

>
> Accepted:
>
> Rejected:
> - *Strict CQRS separation.*

- `python -c "import origo.assets.publish_btc_briefing_feed as m; print(m.publish_btc_briefing_feed.partitions_def is not None)"` returns `True`.
- `python -c "from origo.definitions import defs; print(any(s.name == 'publish_btc_briefing_feed_sensor' for s in defs.sensors))"` returns `True`.
- `pytest tests/origo_source_native/test_publish_btc_briefing_feed.py -q` returns exit code `0`.
- `pytest tests/origo_source_native -q` returns exit code `0`.
- After the first production materialization: `curl -sfL https://huggingface.co/datasets/vaquum/btc_briefing_feed/resolve/main/latest.json | jq -r '.feed_version'` returns `btc_briefing/1`.

### Current State

>
> Accepted:
>
> Rejected:

- `python -c "import origo.assets.publish_btc_briefing_feed as m; print(m.FEED_VERSION)"` returns `btc_briefing/1` (after #295 merges).
- `grep -c huggingface origo/assets/publish_btc_briefing_feed.py` returns `0` (the asset computes and validates; it delivers nowhere).
- `python -c "from origo.definitions import defs; print(any('btc_briefing' in s.name for s in defs.sensors))"` returns `False`.
- `curl -s -o /dev/null -w '%{http_code}' https://huggingface.co/datasets/vaquum/btc_briefing_feed` returns `404`.

### Target State

> predicate.
>
> Accepted:

- `python -c "import origo.assets.publish_btc_briefing_feed as m; print(m.publish_btc_briefing_feed.partitions_def is not None)"` returns `True`.
- `python -c "from origo.definitions import defs; print(any(s.name == 'publish_btc_briefing_feed_sensor' for s in defs.sensors))"` returns `True`.
- `pytest tests/origo_source_native/test_publish_btc_briefing_feed.py -q` returns exit code `0`.
- `pytest tests/origo_source_native -q` returns exit code `0`.
- `pyright origo --outputjson | jq '.summary.errorCount'` does not exceed the value on `main` at the base commit.
- `grep -m1 '^version' pyproject.toml` returns `version = "3.5.0"`.
- `head -1 CHANGELOG.md` returns a line beginning `# v3.5.0`.

### Surfaces

- pyproject.toml
- CHANGELOG.md
- origo/definitions.py
- origo/assets/publish_btc_briefing_feed.py
- tests/origo_source_native/test_publish_btc_briefing_feed.py

Added:
- none

Removed:
- none

### Signature Diff

### origo/assets/publish_btc_briefing_feed.py::publish_btc_briefing_feed
Before:
```python
@asset(name='publish_btc_briefing_feed')
def publish_btc_briefing_feed(context: AssetExecutionContext) -> None:
```
After:
```python
@asset(name='publish_btc_briefing_feed', partitions_def=daily_partitions)
def publish_btc_briefing_feed(context: AssetExecutionContext) -> dict[str, object]:
```

### origo/assets/publish_btc_briefing_feed.py::publish_briefing_feed_to_huggingface
Before:
```
(function does not exist)
```
After:
```python
def publish_briefing_feed_to_huggingface(
    feed: Mapping[str, object], *, repo_id: str, token: str
) -> dict[str, object]:
```

### origo/definitions.py::publish_btc_briefing_feed_sensor
Before:
```
(sensor does not exist)
```
After:
```python
@asset_sensor(
    asset_key=AssetKey('refresh_binance_spot_klines_origo'),
    job=publish_btc_briefing_feed_job,
    default_status=DefaultSensorStatus.RUNNING,
)
def publish_btc_briefing_feed_sensor(
    context: object,
    asset_event: _AssetEventLike,
) -> RunRequest | SkipReason:
```

### Tests That Prove the MVC

|---|---|---|
| Asset is daily-partitioned | tests/origo_source_native/test_publish_btc_briefing_feed.py | test_asset_is_daily_partitioned |
| Sensor is registered and targets the briefing job | tests/origo_source_native/test_publish_btc_briefing_feed.py | test_sensor_is_registered_in_definitions |
| Upload writes the feed file, latest.json and README for the partition day | tests/origo_source_native/test_publish_btc_briefing_feed.py | test_publish_uploads_feed_snapshot_with_latest_pointer |
| latest.json carries feed_version btc_briefing/1 and the day | tests/origo_source_native/test_publish_btc_briefing_feed.py | test_latest_pointer_declares_version_and_day |
| A past day can be published as a partition run | tests/origo_source_native/test_publish_btc_briefing_feed.py | test_backfill_partition_publishes_that_day |

### Regression Gate

>
> Accepted:
>   on non-zero exit.
>
> Rejected:

Workflow file: .github/workflows/pr_checks_tests.yml
Job name: pr_checks_tests
Specific failing command the job runs: `pytest tests/origo_source_native` — which collects `tests/origo_source_native/test_publish_btc_briefing_feed.py` and fails the PR on non-zero exit.

### Out of Scope

- origo/utils/**
- origo/query/**
- origo/assets/sql/**
- .github/workflows/**
- .github/*.json
- docker-compose*.yml
- Dockerfile*
- requirements.txt
- docker-requirements.txt
- tests/tools/**
- tests/fixtures/**

### Risks and Open Questions

Risks:
- The book sections require non-empty depth20 1m coverage for the published day (#295's completeness policy). A day predating the depth capture cannot be backfilled; the sensor path only publishes days the live capture covers, so the daily cadence is unaffected.

Open questions:
- none. Delivery target is `vaquum/btc_briefing_feed` on HuggingFace, token via the existing `HF_TOKEN`/`HUGGINGFACE_HUB_TOKEN` env already present on the production Dagster container (verified during #295). The feed file is the `btc_briefing/1` dict as JSON, one file per day plus a `latest.json` pointer, following the snapshot pattern of `origo/utils/publish_binance_spot_kline_snapshot_to_huggingface.py`. The first sensor tick after deploy publishes the most recent complete UTC day — the "first report" input — and any past covered day is a manual partition launch. Deleting the pre-cutover reports in Vaquum/daily-briefing-btc-usd happens in that repository's cutover change, after the first new-format report exists.

### Closeout

> bot verifies:

Merge SHA:
Merged PR number:
-
-

mikkokotila:

## What does this PR change?

Slice #297, the delivery half of the briefing feed (#295/#296 built and validated it; this PR makes it flow). `publish_btc_briefing_feed` becomes daily-partitioned on the shared spot `daily_partitions`: each partition builds and validates that day's `btc_briefing/1` feed and uploads it to the `vaquum/btc_briefing_feed` HuggingFace dataset via the new `publish_briefing_feed_to_huggingface(feed, *, repo_id, token)` — one `btc_briefing_<yyyymmdd>.json` per day (past days accumulate, nothing is deleted), plus a `latest.json` pointer (`feed_version`, `day`, `file_name`, `sha256`, `generated_at_utc`) and a dataset card. The credential is the existing `HF_TOKEN`/`HUGGINGFACE_HUB_TOKEN` already present on the production Dagster container — no new secret, exactly as decided in #295's Open Questions.

A new `publish_btc_briefing_feed_sensor` (same shape as the twelve kline publish sensors, keyed off `refresh_binance_spot_klines_origo`) fires after each daily spot pipeline materialization, so the first tick after deploy publishes the most recent complete UTC day — the first report's input — and any past day with depth20 coverage is a manual partition launch. The feed dict is JSON-safe by construction since #296 declared every time field as UTC epoch seconds in the SQL.

Five new tests per the slice's Tests table: partitioning, sensor registration, upload contents (feed file byte-round-trips, `latest.json` pointer fields, sha256), and an end-to-end partition materialization with a recording HfApi. Version 3.4.0 → 3.5.0.

## Checklist

- [x] I ran the project's test suite locally without errors (where applicable) — `pytest tests/origo_source_native/test_publish_btc_briefing_feed.py -q`: 18 passed; full suite run reported in a follow-up comment; `pyright` (1.1.408): 0 errors in the changed modules, total unchanged from `main`

Closes #297


mikkokotila:
Full `pytest tests/origo_source_native` run on this head locally: 183 passed (the 178 from #296 plus the five new tests from #297's Tests table). `pyright` 1.1.408: total error count unchanged from `main`, 0 errors in the changed modules. Both MVC probes return `True`: `publish_btc_briefing_feed.partitions_def is not None` and `publish_btc_briefing_feed_sensor` in `defs.sensors`.

origo/assets/publish_btc_briefing_feed.py
```diff
@@ -196,24 +206,117 @@ def build_briefing_feed(
...
+    Writes ``btc_briefing_<yyyymmdd>.json``, a ``latest.json`` pointer and the
+    dataset card, then uploads the folder in one commit. Past day files are
+    left in place, so the dataset accumulates one file per published day.
+    """
+    day = feed['day']
+    if not isinstance(day, str):
+        raise RuntimeError(f'Feed day must be an ISO date string, got {day!r}.')
+
+    file_name = f'btc_briefing_{day.replace("-", "")}.json'
+    feed_bytes = (json.dumps(feed, sort_keys=True) + '\n').encode('utf-8')
```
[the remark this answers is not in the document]
mikkokotila: Fixed in 1543df0: the day is now parsed with `date.fromisoformat` before any use — a non-date value raises instead of reaching the filesystem — and both the file name and the published `day` derive from the parsed, normalized value, so no unvalidated string can shape the path or the `latest.json` pointer.

origo/assets/publish_btc_briefing_feed.py
```diff
@@ -196,24 +206,117 @@ def build_briefing_feed(
...
+    sha256 = hashlib.sha256(feed_bytes).hexdigest()
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp_path = Path(tmp_dir)
+        (tmp_path / file_name).write_bytes(feed_bytes)
+        (tmp_path / 'latest.json').write_text(
```
bit-mis: p1: `latest.json` and the README dataset card are rewritten unconditionally with whichever partition day the run publishes, so materializing an older partition (the documented way to publish a past day with depth20 coverage, now that the asset is daily-partitioned) silently repoints the published `latest` pointer and card at a stale day, and when several partitions are in flight the surviving pointer is just whichever run committed last rather than the newest day present in the dataset. The per-day feed files are safe because each carries its own name, but consumers that follow `latest.json` to find the newest feed get older data than the dataset already contains; `test_backfill_partition_publishes_that_day` cannot catch this because it publishes only a single day, so no newer pointer ever exists to be clobbered.

bit-mis (changes requested):

origo/assets/publish_btc_briefing_feed.py
mikkokotila: Fixed in fe7a1bd: publish now reads the dataset's current latest.json first (via repo_exists/file_exists/hf_hub_download, so an absent repo or pointer means first publish) and rewrites the pointer and dataset card only when this run's day is strictly newer — a backfilled older partition uploads its per-day file and leaves latest untouched (`latest_updated: False` in the returned metadata). `test_backfill_of_older_day_does_not_move_latest_pointer` pins the skip path against an existing newer pointer, `test_newer_day_advances_latest_pointer` the advance path. Residual race honestly stated: two forward-publishing runs in flight simultaneously can still interleave check-then-write and leave the pointer one day behind until the next publish heals it — the sensor launches one run per daily materialization, so in practice the monotonic guard closes the reported hole.

origo/definitions.py
```diff
@@ -1230,6 +1230,21 @@ def publish_binance_spot_240M_dollar_klines_to_huggingface_sensor(
...
+@asset_sensor(
+    asset_key=AssetKey("refresh_binance_spot_klines_origo"),
+    job=publish_btc_briefing_feed_job,
+    default_status=DefaultSensorStatus.RUNNING,
+)
+def publish_btc_briefing_feed_sensor(
+    context: object,
+    asset_event: _AssetEventLike,
+) -> RunRequest | SkipReason:
+    return _publish_binance_spot_klines_to_hf_run_request(
```
bit-mis, line 1242: p0: this sensor delegates to `_publish_binance_spot_klines_to_hf_run_request`, a helper shared with the twelve non-partitioned kline publishers that carries only a run key and no partition key, so every automatic tick launches a partition-less run of the now daily-partitioned `publish_btc_briefing_feed_job` and dies on `context.partition_key`, meaning the daily cadence this PR exists to establish never publishes a feed.

origo/assets/publish_btc_briefing_feed.py
```diff
@@ -196,24 +206,173 @@ def build_briefing_feed(
...
+    published_day = date.fromisoformat(raw_day)
+    day = published_day.isoformat()
+
+    file_name = f'btc_briefing_{day.replace("-", "")}.json'
+    feed_bytes = (json.dumps(feed, sort_keys=True) + '\n').encode('utf-8')
+    sha256 = hashlib.sha256(feed_bytes).hexdigest()
+
+    api = _make_hf_api(token)
+    latest_day = _published_latest_day(api, repo_id)
+    advance_latest = latest_day is None or published_day > latest_day
```
bit-mis: p2: the pointer advances only on a strictly newer day, so re-publishing the same day (partition re-materialization after a failed upload or late-arriving data) overwrites `btc_briefing_<day>.json` with new bytes while `latest.json` keeps the sha256 of the previous bytes, leaving consumers that verify the checksum of the file named by the pointer rejecting the current feed as corrupt.

bit-mis (changes requested):

mikkokotila:
Full `pytest tests/origo_source_native` on fe7a1bd locally: 185 passed (adds the latest-pointer skip and advance tests). `pyright` 1.1.408 total unchanged from `main`; all CI gates green on this head.

origo/definitions.py
mikkokotila: Not addressed, because the premise does not hold: `_publish_binance_spot_klines_to_hf_run_request` forwards the partition key from the materialization event — definitions.py lines 1023-1029 read `partition_key = asset_event.dagster_event.partition`, skip when it is `None`, and return `RunRequest(partition_key=partition_key, run_key=f"{run_key_prefix}::{partition_key}")`. The twelve kline publishers this helper already serves are themselves daily-partitioned on the same `daily_partitions` (e.g. publish_binance_spot_15m_klines_to_huggingface.py line 10), so the helper was built precisely for partitioned target jobs, and every automatic tick launches a partitioned run whose `context.partition_key` is the materialized day. `test_backfill_partition_publishes_that_day` materializes the asset with a partition key through the same job path and passes.

origo/assets/publish_btc_briefing_feed.py
mikkokotila: Fixed in 07cda47: the pointer now advances on `>=` instead of `>`, so re-publishing the pointed-at day rewrites latest.json alongside the overwritten per-day file and the pointer's sha256 always matches the current bytes of the file it names; backfilling a strictly older day still never moves the pointer. `test_same_day_republish_refreshes_latest_pointer` pins the refresh, and the existing skip/advance tests still pass.

bit-mis (approved):

mikkokotila:
Full `pytest tests/origo_source_native` on 07cda47 locally: 186 passed (adds the same-day pointer-refresh test). `pyright` 1.1.408 total unchanged from `main`. All CI gates green on this head.

